PR #79, opened 2023-05-22
[#38] [Backend] As a user I can login with an OAuth provider

### LINKED ISSUE #38: [Backend] As a user I can login with an OAuth provider
## Why

Integrating a third-party service is an oft-required basic need. In addition, implementing an OAuth authentication provider can make building a POC simpler.

## Acceptance Criteria

- Set up the third-party library required for Google login. Ideally, Passport should be used instead of NextAuth.
- Create a new table to store users.
- Implement the model, controller, and service layers for the login.
- Implement logic for protected and public routes, i.e., return a 403 error page when accessing a protected route.

tednguyendev:
Close #38

## What happened 👀
- Add API endpoint to login with Google
- Store session in the cookie
- Store user in the database
- Use the [New app Directory (Beta)](https://nextjs.org/blog/next-13#new-app-directory-beta) of Next.js 13
- Use [Route Handler](https://nextjs.org/docs/app/building-your-application/routing/router-handlers#api-reference) of Next.js 13
- Use [NextAuth](https://next-auth.js.org/) to oAuth

### Reuse old structure Next.js code
This PR mostly reuse the code from PR #2 for convenience.

### NextAuth
When compare [NextAuth](https://next-auth.js.org/) to [Passport](https://www.passportjs.org/), we can easily see that Passport is more battle-tested, support more oAuth providers, more popular([21.5k stars on Github](https://github.com/jaredhanson/passport), compared to [16.7k of NextAuth](https://github.com/nextauthjs/next-auth)), but I see NextAuth is more suitable for our POC, since:
- It's build for Next.js, hence more compatible and always keep up-to-date with Next.js, and it's [compatible with `Route Handler`](https://next-auth.js.org/configuration/initialization#route-handlers-app)
  - Meanwhile, seems like the composition `Passport` and `Route Handler` and `next-connect` are not [compatible yet](https://github.com/hoangvvo/next-connect/issues/220#issuecomment-1520623909). And it's also because the `Route Handler`of Next.js 13 can already replace some parts of the `next-connect` now.
- A lot easier to setup oAuth, which only needs to [setup envs](https://github.com/nimblehq/js-framework-benchmark/pull/79/files#diff-795b86af04c107fdf257bbc42f5724fea71ba2b6e7cfd39ca8df44415df5e68bR4-R5), setup the [provider credentials](https://github.com/nimblehq/js-framework-benchmark/pull/79/files#diff-d4019c1675319debdd9edc94a7be90680712c40c9a4ccedc2789423592bd422bR1-R20), and a [`route.js` file](https://github.com/nimblehq/js-framework-benchmark/pull/79/files#diff-700197f1dcfbea6d933ae8194a9d3d66879c96279b26527c33cb04bb399d0cdcR1-R3),
- [Easy to set up middleware](https://github.com/nimblehq/js-framework-benchmark/pull/79/files#diff-8dc469b9cf60785ce08e724cab36052f5f6278de35da8aea179287ddbe5c6ac3R1)

## Proof Of Work 📹
- Login flow: 

https://github.com/nimblehq/js-framework-benchmark/assets/76901736/907a3818-cc1a-4052-a309-ddd225539531

- Stored Cookie:

CleanShot 2023-05-23 at 15 18 57@2x


nextjs-13/package.json
```diff
@@ -6,15 +6,31 @@
...
     "@types/react": "18.2.6",
     "@types/react-dom": "18.2.4",
     "eslint": "8.40.0",
     "eslint-config-next": "13.4.3",
+    "iron-session": "^6.3.1",
+    "jest": "^29.5.0",
+    "jest-environment-jsdom": "^29.5.0",
+    "jest-mock-extended": "^3.0.4",
     "next": "13.4.3",
+    "next-connect": "0.13.0",
```
olivierobert, line 30: This package is now outdated as there is a newer 1.0.0 version that has been in the works and released: https://github.com/hoangvvo/next-connect/releases/tag/v1.0.0

olivierobert (commented):
While re-using existing code is a good idea, I am unsure if we should still keep using `next-connect` with Next.JS 13. 

I used that package as 1) passport is THE battle-tested oAuth library in the Node community, and 2) API routes in Next.JS 12 (with the Pages router) were quite "poor" (we had to check on the HTTP verb, e.g., `if req == 'POST'). 

However, Next.JS 13 (with the App router) has better server-side components, including [routes handlers](https://nextjs.org/docs/app/building-your-application/routing/router-handlers). So we can do defining GET and POST endpoints with  `function GET() {}` and `function POST() {}`. So I believe we can do more backend logic similarly to what we do with Express (see [this article](https://codevoweb.com/google-oauth-authentication-react-and-node/) to do oAuth without even passport). 

I am not against using https://next-auth.js.org/ if it is easier (e.g, easier to test) and better (e.g., more secure, bug-free). I want to make sure we understand our options.

tednguyendev:
> While re-using existing code is a good idea, I am unsure if we should still keep using `next-connect` with Next.JS 13.
> 
> I used that package as 1) passport is THE battle-tested oAuth library in the Node community, and 2) API routes in Next.JS 12 (with the Pages router) were quite "poor" (we had to check on the HTTP verb, e.g., `if req == 'POST').
> 
> However, Next.JS 13 (with the App router) has better server-side components, including [routes handlers](https://nextjs.org/docs/app/building-your-application/routing/router-handlers). So we can do defining GET and POST endpoints with `function GET() {}` and `function POST() {}`. So I believe we can do more backend logic similarly to what we do with Express (see [this article](https://codevoweb.com/google-oauth-authentication-react-and-node/) to do oAuth without even passport).
> 
> I am not against using https://next-auth.js.org/ if it is easier (e.g, easier to test) and better (e.g., more secure, bug-free). I want to make sure we understand our options.

Yes I see.

I just updated the PR description. What do you think about this?

And there's also this one thing. When using passport authenticate, we are using it as `authService.authenticate()(req, res)`, which needs both request and response, but since the syntax of Route Handlers is `export async function GET(request: Request) {}`, we don't have the response here. Do you know if there are any other ways for us to do this? The Nextjs and it's ecosystem are still new to me so I appreciate your help a lot:D.

tednguyendev:
> @tednguyendev The primary issue is that this implementation uses an outdated version of `next-connect`. That will not do. We should always use the latest version of libraries for security and maintenance purposes. The secondary issue is that this POC does not leverage the latest features of Next.JS 13, namely the App Router. That will also not do since the goal of the POC is to understand if Next.JS can be a good solution to build full-stack applications.
> 
> So you need to revisit the implementation to address the above concerns: 1) not using an outdated library, and 2) using the new App router. You can try to do it with Passport first and if you cannot make it work then fall back to NextAuth.

Got it! I will reimplement this and will update you later. Thank you! 😁

nextjs-13/.env.sample
phisith, line 1: I think you are missing EOF line break

phisith (approved):

nextjs-13/.env.sample
tednguyendev: If I'm not wrong, I remember that .env files do not need EOF 😄

Whenever a file should have a EOF and is missing the EOF, Github usually show this icon to reminder the authors to add the EOF:

image

phisith: actually this morning Oliver also requested me to add EOF line break on my .env.sample file.

not sure If we both wrong 😂

https://github.com/nimblehq/js-framework-benchmark/pull/80#discussion_r1203319242

tednguyendev: Ah 🤣 What do you think about this @olivierobert?

I then just searched and asked ChatGPT for sources about this convention but seems like no one actually talked about it.

ChatGPT just said:

> However, the lack of a specific end-of-file line is a common characteristic of .env file formats.

And if looking at the .env examples in [this guideline](https://github.com/motdotla/dotenv#-examples), it may seem like .env files truly usually not have the EOF newline, example are [this](https://github.com/dotenv-org/examples/blob/master/dotenv-nodejs/.env), [this](https://github.com/dotenv-org/examples/blob/master/dotenv-typescript/.env), and [this](https://github.com/dotenv-org/examples/blob/master/dotenv-express/.env).

I'm just curious haha, so I'm still super okay if you think we should have the EOF newline.

olivierobert: I usually rely on the code editor to add these automatically. I only raise issues when it shows as missing on GitHub. So not having it seems okay 👍

nextjs-13/prisma/migrations/migration_lock.toml
```diff
@@ -0,0 +1,3 @@
+# Please do not edit this file manually
+# It should be added in your version-control system (i.e. Git)
+provider = "postgresql"
```
olivierobert, line 3: Related to [this discussion](https://github.com/nimblehq/js-framework-benchmark/pull/79/files#r1203377631), for instance, this file shows that is missing an EOF line is missing

olivierobert (approved):
Looks great overall 👍 Just two minor issues

nextjs-13/src/repositories/user.repository.test.ts
```diff
@@ -0,0 +1,67 @@
...
+        const user = { ...userFactory, ...userAttributes };
+
+        dbClientMock.user.findUnique.mockResolvedValue(user);
+
+        await expect(findUserById('1')).resolves.toEqual(user);
+      });
+    });
+
+    describe('given there is NO matching user', () => {
+      it('return null', async () => {
```
olivierobert, line 22: Minor typo issue:

```suggestion
      it('returns null', async () => {
```

nextjs-13/prisma/migrations/migration_lock.toml
tednguyendev: Ahh, this is the file `prisma` automatically creates when I use `prisma generate`. Ok let me add the EOF :D

tednguyendev: Oke done!

nextjs-13/src/repositories/user.repository.test.ts
tednguyendev: I just fixed it!

olivierobert (approved):

PR #80, opened 2023-05-22
[#56] Create GitHub action CI workflow

phisith:
https://github.com/nimblehq/js-framework-benchmark/issues/56
## What happened 👀

- Add Unit Testing
- Setup CI

## Insight 📝

Using `Jest` and `RTL (React Testing Library)` for Unit Testing.
Setup Github action runner and active when pull request.

## Proof Of Work 📹
Screenshot 2566-05-22 at 15 55 55

PR #81, opened 2023-05-23
[#57] Backend as a user I can login with an Oauth provider

phisith:
https://github.com/nimblehq/js-framework-benchmark/issues/57

## What happened 👀

- Setup authentication (login using Google), 
- Save user information into database when login
- Add login url and protected route

## Insight 📝

To be able to setup the authentication I'm using the [remix-auth](https://www.npmjs.com/package/remix-auth) with [remix-auth-socials](https://www.npmjs.com/package/remix-auth-socials) (A collection of Remix Auth strategies for Oauth2 Social logins.).

In this setup I'm not able to use the `Passport.js`, due to Remix doesn't have the ability to control the server code like Nextjs. 

Reference:
Screenshot 2566-05-23 at 17 58 36

### Login Flow
https://github.com/nimblehq/js-framework-benchmark/assets/59072804/4f705730-4e75-462b-9168-73e26f2f49cf

### session is store in cookies
Screenshot 2566-05-24 at 09 22 03

### Protected route (must login before access the route)
Screenshot 2566-05-23 at 16 04 54


tednguyendev (approved):

remix/app/routes/newsletters.tsx
```diff
@@ -0,0 +1,61 @@
+import type { User } from "@prisma/client";
+import type { LoaderArgs } from "@remix-run/node";
+import {
+  Form,
+  isRouteErrorResponse,
+  useLoaderData,
+  useRouteError,
+} from "@remix-run/react";
+import { SocialsProvider } from "remix-auth-socials";
```
olivierobert, line 9: Let's ada line break between third-party (node module) imports and application imports

```suggestion
import { SocialsProvider } from "remix-auth-socials";

```

remix/app/utils/db.server.ts
olivierobert, line 1: `utils` tend to contain too many unrelated pieces of code over time. How about moving this into `app/config` instead?

remix/app/services/auth.server.ts
```diff
@@ -0,0 +1,44 @@
+import { Authenticator } from "remix-auth";
+import { sessionStorage } from "./session.server";
+import { GoogleStrategy, SocialsProvider } from "remix-auth-socials";
+import type { UserProfile } from "~/types";
+import { db } from "~/utils/db.server";
+
+export const authenticator = new Authenticator(sessionStorage);
+
+async function handleSocialAuthCallBack({ profile }: any) {
+  const userProfile: UserProfile = profile?._json;
```
olivierobert, line 10: Add a line breaks between variable definition and code logic locks for readability:

```suggestion
  const userProfile: UserProfile = profile?._json;
  
  if (userProfile) {
```

remix/app/services/auth.server.ts
```diff
@@ -0,0 +1,44 @@
...
+        },
+      });
+    } catch (err) {
+      console.log(err);
+    }
+  }
+  return profile;
+}
+
+authenticator.use(
```
olivierobert, line 34: Shouldn't this be initialized in a config and upon booting the application? If no credentials are found, the app should not start.

remix/app/services/session.server.ts
```diff
@@ -0,0 +1,14 @@
+import { createCookieSessionStorage } from "@remix-run/node";
+
+export const sessionStorage = createCookieSessionStorage({
+  cookie: {
+    name: "__session",
+    sameSite: "lax",
+    path: "/",
+    httpOnly: true,
+    secrets: ["veryscret"],
```
olivierobert, line 9: This should define as an env variable

remix/app/routes/newsletters.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const { user } = useLoaderData();
+  return (
+    <div>
+      <h1>Newsletter Page</h1>
+      <p>{user?.name}</p>
+    </div>
+  );
+}
+
+export function ErrorBoundary() {
```
olivierobert, line 42: This error boundary should be define at the layout level to ensure this code is not repeated. See https://dev.to/pckilgore/remix-sharing-your-index-layout-with-sub-routes-2856

remix/app/routes/newsletters.tsx
```diff
@@ -0,0 +1,61 @@
...
+    throw new Response("Forbidden", { status: 403 });
+  }
+  let user: User | null = null;
+
+  try {
+    user = await db.user.findUnique({
+      where: { email: userProfile.email },
+    });
+  } catch (err) {
+    console.log(err);
```
olivierobert, line 27: Let's avoid using `console.log` (as it is unsafe and we tend to forget removing it) and instead use `throw` so that it can be caught by the `ErrorBoundary`

remix/app/routes/newsletters.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { authenticator } from "~/services/auth.server";
+import type { UserProfile } from "~/types";
+import { db } from "~/utils/db.server";
+
+export const loader = async ({ request }: LoaderArgs) => {
+  const userRequest = (await authenticator.isAuthenticated(request)) as any;
+  const userProfile: UserProfile = userRequest?._json;
+  if (!userProfile) {
+    throw new Response("Forbidden", { status: 403 });
+  }
```
olivierobert, line 19: Avoid mixing variable definitions and business logic. Any if/else should be placed after variables definitions (unless you work with Go which has a very procedural approach 😉 )

```suggestion
  const userRequest = (await authenticator.isAuthenticated(request)) as any;
  const userProfile: UserProfile = userRequest?._json;
  let user: User | null = null;
  
  if (!userProfile) {
    throw new Response("Forbidden", { status: 403 });
  }
```

remix/app/routes/newsletters.tsx
```diff
@@ -0,0 +1,61 @@
...
+  let user: User | null = null;
+
+  try {
+    user = await db.user.findUnique({
```
olivierobert, line 23: For DB operations, it is best to use the repository pattern by defining a new class for each model/entity:

```
// in app/repositories
import { db } from `~/config/db.server";

class UserRespository {

  static function findBy(param): User {
     //....
  }

  // other functions
}
```

It would allow to `) not having to import `db` in any routes (so abstracting where data is coming) and implement adequate typing/

olivierobert (changes requested):
This branch should not have the code changes from #80 to make the code review focused on the changes 🙈 In addition, skipping tests will not do.

remix/app/routes/newsletters.tsx
olivierobert: Also shall we use https://remix.run/docs/en/main/route/error-boundary-v2?

remix/app/utils/db.server.ts
phisith: that's sound better idea, Thank you 🙏🏼

remix/app/services/auth.server.ts
phisith: `authenticator.use(` yes, that was my mistake it should initialized in the config, at the time when I initialized it I was focus on how to make work. so now I will move it to config.

remix/app/routes/newsletters.tsx
phisith: Thank you for suggestion and the link, but it seem like don't work anymore with Remix route v2.

To avoid the repeat error boundary I have move the it to `root.tsx`.

For the next implementation I will follow the [Remix route v2](https://remix.run/docs/en/main/file-conventions/route-files-v2#nested-layouts-without-nested-urls)

remix/app/routes/newsletters.tsx
phisith: Yes the error boundary I have already follow the V2 version 😄.

phisith:
Hi @olivierobert, I have resolved all of the requested in today. the only thing that I have left is testing 😅, I'm working on now.

Thank you 🙏🏼

remix/Makefile
```diff
@@ -0,0 +1,18 @@
+.PHONY: up start_db
+
+up:
```
olivierobert, line 3: I recommend renaming the commans to `start` and `stop` as it makes more sense to start and stop and application 😅

remix/app/config/db.server.ts
olivierobert, line 1: Always prefer naming files, functions, classes, etc. with their full name: `database.server.ts`. Clarity over brevity.

remix/app/config/db.server.ts
```diff
@@ -0,0 +1,19 @@
+import { PrismaClient } from "@prisma/client";
+
+let db: PrismaClient;
```
olivierobert, line 3: That code below looks complex. Do we need it all? Can we shorten it to:

```javascript
import { PrismaClient } from '@prisma/client';

const dbClient = new PrismaClient();

export default dbClient;
```

remix/app/repositories/user.server.test.ts
```diff
@@ -0,0 +1,55 @@
+import type { User } from "@prisma/client";
+import { prismaMock } from "../tests/setup-test-env";
+import UserRespository from "../repositories/user.server"
+
+describe("Test user model", () => {
```
olivierobert, line 5: This should describing the object under test:

```suggestion
describe("User Repository", () => {
```

remix/app/repositories/user.server.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    test("Create User", async () => {
```
olivierobert, line 6: This should be the method name:

```suggestion
    test("updateOrCreate", async () => {
```

remix/app/repositories/user.server.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        const user: User = {
+            name: "Beckham",
```
olivierobert, line 8: Creating all these attributes by hand will get tiring and repetitive. It is best to use factories for text pictures: https://github.com/nimblehq/js-framework-benchmark/blob/develop/nextjs-12/test/factories/user.factory.ts

remix/app/repositories/user.server.test.ts
```diff
@@ -0,0 +1,55 @@
...
+            email: "phisith@nimblehq.co",
+            avatarUrl: "weofkwepofkwefokwepfokwpo",
+            createdAt: new Date(),
+            id: 1
```
olivierobert, line 12: Shouldn't this ID be created automatically?

remix/app/repositories/user.server.ts
```diff
@@ -0,0 +1,39 @@
...
+                    avatarUrl: data.picture,
+                },
+            });
+        } catch (err) {
+            throw err;
+        }
+    }
+}
+
+export default UserRespository;
```
olivierobert, line 39: Typo detected:

```suggestion
export default UserRepository;
```

remix/app/routes/auth.google.callback.tsx
```diff
@@ -0,0 +1,11 @@
+import type { LoaderArgs } from "@remix-run/node";
+import { authenticator } from "~/config/auth.server";
+
+import { SocialsProvider } from "remix-auth-socials";
```
olivierobert, line 4: The ordering of imporst should be (third-party first and application code next):

```suggestion
import type { LoaderArgs } from "@remix-run/node";
import { SocialsProvider } from "remix-auth-socials";

import { authenticator } from "~/config/auth.server";
```

remix/app/tests/setup-test-env.ts
```diff
@@ -0,0 +1,16 @@
...
+jest.mock("../config/db.server", () => ({
+    __esModule: true,
+    default: mockDeep<PrismaClient>(),
+}))
+
+beforeEach(() => {
+    mockReset(prismaMock)
+})
+
+export const prismaMock = db as unknown as DeepMockProxy<PrismaClient>
```
olivierobert, line 16: ANother EOF lien break missing

remix/app/services/auth.server.ts
```diff
@@ -0,0 +1,11 @@
+import type { UserProfile } from "~/types";
+import UserRespository from "~/repositories/user.server";
+
+export async function handleSocialAuthCallBack({ profile }: any) {
+  const userProfile: UserProfile = profile?._json;
+
+  if (userProfile) {
+    await UserRespository.updateOrCreate(userProfile)
+  }
```
olivierobert, line 9: Add also a line break before the return statements. Always consider breaking code into: variables, logic, and return values.

olivierobert (changes requested):

phisith:
Today I have rebase develop into this branch and add the code format (Eslint), I will working on the request change tomorrow.

remix/Makefile
phisith: Ok, I have fixed it

remix/app/config/db.server.ts
phisith: Well noted 👌🏼

remix/app/config/db.server.ts
phisith: My code below is come from Remix suggestion that I follow them at the beginning, but I think follow your way (that also suggestion from prisma) I think is better.

Fixed

remix/app/repositories/user.server.test.ts
phisith: Fixed.

remix/app/repositories/user.server.test.ts
phisith: Fixed.

remix/app/repositories/user.server.test.ts
phisith: Yes, my prisma will auto generate the id.  For the reason that I declare a id here is because Typescript is force me to do that, I think to solve this issues is better to Generate Mockup data Library like `Faker.js`, like you have mention above.

phisith: Fixed.

PR #91, opened 2023-05-26
[#58] [Frontend] As a user I can login with an Oauth provider

phisith:
[#58]

## What happened 👀

- Implement UI to the authentication
- Create Component and Share Layout for public and protected route
- Improve routing in the app
- Add Testing for the component

## Insight 📝

For the testing I'm able to use add testing to most of the components and other functions, but there are some Issues that make testing Remix isn't a good experience, like If I set the jest `testEnvironment` to `jsdom` I will be able to test those components and functions that import or relate to `@Remix/node`, on the other side if I change the testEnvironment to `node`, I will be able to test the that import or relate `@Remix/node` but can't test the components and functions relate to `React` components, it mean you have choose between testing the `node` function or `any react` and `route` components. 

After I have done some research I found that Testing strategy is still in Remix is discussion, but for the reliable one it would be using `Cypress` for testing whole route E2E and `Jest` for the components and other small part. [ 👉🏼 Reference](https://github.com/remix-run/remix/discussions/993)

Not sure, If we need Cypress for testing now, but we can discuss below.

In my opinion, Remix is in the transitioning to V2 some of the resource that I found some work or not work and the worst case is just outdated, so mostly I will stick with the official doc but there is no information about testing yet 😞.

## Proof Of Work 📹

https://github.com/nimblehq/js-framework-benchmark/assets/59072804/a6306c6b-5324-4a8f-88fc-74369c68df78


tednguyendev:
I had the same issue with `testEnvironment` in my Next.js project too.

I found a way to deal with it is to use `jest-environment-jsdom` as the default `testEnvironment`, and in the tests that can not use `jest-environment-jsdom`, I add [these lines](https://github.com/nimblehq/js-framework-benchmark/pull/94/files#diff-2ee6b5c4cbf359c8cda60ecacf45024db9a86bbe0eaa6ad5d51dfb7ee9780aa4R1-R3) to them.

Not the best way but it works 😁.

phisith:
I will try to add this to my test file and see if it work. 
Thank you 😆!

remix/app/repositories/user.server.ts
```diff
@@ -0,0 +1,39 @@
...
+import dbClient from '../config/database.server';
+import type { UserProfile } from '../types';
+
+class UserRepository {
+  static async findBy(param: UserProfile): Promise<User> {
+    try {
+      return (await dbClient.user.findUnique({
+        where: param,
+      })) as User;
+    } catch (err) {
```
tednguyendev, line 12: I think we can remove the `try` + `catch` + `throw` here, since the catch only re-throw the error 🤔

tednguyendev: The same may applies to the `updateOrCreate` 👀

tednguyendev (approved):

phisith:
> I had the same issue with `testEnvironment` in my Next.js project too.
> 
> I found a way to deal with it is to use `jest-environment-jsdom` as the default `testEnvironment`, and in the tests that can not use `jest-environment-jsdom`, I add [these lines](https://github.com/nimblehq/js-framework-benchmark/pull/94/files#diff-2ee6b5c4cbf359c8cda60ecacf45024db9a86bbe0eaa6ad5d51dfb7ee9780aa4R1-R3) to them.
> 
> Not the best way but it works 😁.

phisith:
> I had the same issue with `testEnvironment` in my Next.js project too.
> 
> I found a way to deal with it is to use `jest-environment-jsdom` as the default `testEnvironment`, and in the tests that can not use `jest-environment-jsdom`, I add [these lines](https://github.com/nimblehq/js-framework-benchmark/pull/94/files#diff-2ee6b5c4cbf359c8cda60ecacf45024db9a86bbe0eaa6ad5d51dfb7ee9780aa4R1-R3) to them.
> 
> Not the best way but it works 😁.

Oh, @Ted Suggestion is work, Thanks a lot for that. I spend a night find a solution, but seem no need talking about this Remix community 🙉

remix/app/repositories/user.server.ts
phisith: Can you a bit more about re-throw ?, because in my app in `root.tsx` has Error boundary that I try catch to throw the error to call the Error boundary.

tednguyendev: Hmm 🤔 Let me take a look at this

tednguyendev: What I meant in the above just it's that instead of doing this:

```typescript
static async findBy(param: UserProfile): Promise<User> {
  try {
    return (await dbClient.user.findUnique({
      where: param,
    })) as User;
  } catch (err) {
    throw err;
  }
}
```

We can instead make it more concise like this 😆:
```typescript
static async findBy(param: UserProfile): Promise<User> {
  return (await dbClient.user.findUnique({
    where: param,
  })) as User;
}
```

Does doing this impact the behavior of your Error boundary? 🤔

phisith: Oh, now I understand what you mean re-throw, actually we don't need to use `try` +`catch` +`throw` because `prisma` it throw error by itself. Thanks for trying to help me on this 😅🙏🏼. (I will do the change)

phisith: • [65182a7](https://github.com/nimblehq/js-framework-benchmark/pull/91/commits/65182a70a4c14597e4027c744c4ccaaadd4132f3) Fixed here.

olivierobert:
@phisith this PR needs to be rebased since #90 was merged. Time to practice rebasing again.

phisith:
Hi @olivierobert for fix the conflicting files here we prefer to use `git rebase` right ?

phisith:
@olivierobert Currently I mess up with git rebase again now @tednguyendev helping me on that 🙉

olivierobert (commented):
Just some minor adjustments and then this can be merged

remix/app/services/auth.server.ts
```diff
@@ -1,18 +1,18 @@
...
 }
 
 export async function handleSocialAuthCallBack({ profile }: Profile) {
   const userProfile: UserProfile = profile?._json;
 
   if (userProfile) {
-    await UserRepository.updateOrCreate(profile?._json as User);
+    await UserRepository.updateOrCreate(userProfile);
+  } else {
+    throw new Response('This action need authenticated', { status: 401 });
```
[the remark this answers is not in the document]
phisith: Well noted. Solved on: [029f36c](https://github.com/nimblehq/js-framework-benchmark/pull/91/commits/029f36c39c4bb0a841d377459057d77423ad752c)

remix/app/routes/auth.sign-in.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const session = await middleware(request);
+
+  return { session };
+}
+
+export default function AuthSignInPage() {
+  return (
+    <>
+      <section className="flex flex-col gap-4">
+        <article>
```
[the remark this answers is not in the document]
phisith: Solved on: [488da9a](https://github.com/nimblehq/js-framework-benchmark/pull/91/commits/488da9abb965e2956d27ebb318f63a31162ee51c)

olivierobert (approved):

PR #96, opened 2023-05-29
[#44] [Backend] As a user I can create a newsletter

### LINKED ISSUE #44: [Backend] As a user I can create a newsletter
## Why

Users can add new records of newsletters as needed.

## Acceptance Criteria

- Add a new route to create newsletters.
- Create the controller and service layers to create newsletters.
- Ensure validation rules are implemented.

tednguyendev:
Close #44 

## What happened 👀
- Add new `newsletters` table
- Add new route to create `newsletter`
- Add [`baseHandler`](https://github.com/nimblehq/js-framework-benchmark/pull/96/files#diff-add2a84e0fcd856a4db7e6914984152b0d780f178b1988d446d1d0d0a9e7eac6R1) to DRY the code to authenticate users whenever an API is called
- Add `updatedAt` to `users` also
- Add corresponding tests

### `Prisma` error format
Personally, I don't like the way prisma support us at handling exceptions.

While the errors on the server logs for developers to read are very detailed, Prisma seems to not support formatting the error details to the client side([example](https://github.com/prisma/prisma/issues/9082#issuecomment-1175304006.)). Seems like the [standard way](https://www.prisma.io/docs/concepts/components/prisma-client/handling-exceptions-and-errors) to handle database errors is to just if else a lot of times to check for error codes then return the error message ourselves, like this, or [this](https://stackoverflow.com/a/75600318):

```javascript
import { PrismaClient, Prisma } from '@prisma/client'

const client = new PrismaClient()

try {
  await client.user.create({ data: { email: 'alreadyexisting@mail.com' } })
} catch (e) {
  if (e instanceof Prisma.PrismaClientKnownRequestError) {
    // The .code property can be accessed in a type-safe manner
    if (e.code === 'P2002') {
      console.log(
        'There is a unique constraint violation, a new user cannot be created with this email'
      )
    }
  }
  throw e
}
```

To keep it simple, for now, I will just implement basic validations using [this if else statement](https://github.com/nimblehq/js-framework-benchmark/pull/96/files#diff-2b438c9d13a48fc01fdc6132036acd160bdbd4b65f4cce429765a9e7e2c811feR19). If we need detailed error information for the frontend to handle, since `Prisma` not support this as good as `Active Record`, I suggest we use the schema validators like [joi](https://github.com/hapijs/joi) instead.

## Proof Of Work 📹
CleanShot 2023-06-01 at 13 56 48@2x


nextjs-13/src/app/api/v1/me/route.test.ts
```diff
@@ -4,58 +4,37 @@
...
 import { User } from '@prisma/client';
 import { StatusCodes } from 'http-status-codes';
-import * as nextAuthJwtModule from 'next-auth/jwt';
 
 import { dbClientMock } from '@test/database';
 import { userFactory } from '@test/factories/user.factory';
+import baseHandler from 'lib/handler/base.handler';
 
 import { GET } from './route';
+jest.mock('lib/handler/base.handler');
```
phisith, line 13: I think you should I add the line-break between the `import` and `jest.mock`

nextjs-13/src/app/api/v1/newsletter/route.test.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * @jest-environment node
+ */
+
+import { Newsletter, Prisma } from '@prisma/client';
+
+import { dbClientMock } from '@test/database';
+import { newsletterFactory } from '@test/factories/newsletter.factory';
+import baseHandler from 'lib/handler/base.handler';
+
```
phisith, line 10: I guess just remove this line break since those are the application import

nextjs-13/src/lib/handler/base.handler.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    if (token && token.userId) {
+      const currentUser = await findUserById(token.userId as string);
+
+      if (currentUser) {
+        let body = {};
+
+        try {
+          body = await req.json();
+        } catch (err) {}
```
phisith, line 22: Not sure why this catch is empty 😅

nextjs-13/src/lib/handler/base.handler.ts
```diff
@@ -0,0 +1,37 @@
...
+
+        console.log('========>body : ', body);
```
phisith, line 24: Maybe we shouldn't leave the console.log in the code base here, lets confirm it with @olivierobert I guess.

nextjs-13/src/lib/handler/base.handler.ts
```diff
@@ -0,0 +1,37 @@
...
+        return await callback(currentUser, body);
+      }
+    }
+
+    return NextResponse.json({ message: 'Invalid token' }, { status: 401 });
+  } catch (err) {
+    console.log('========>err : ', err);
```
phisith, line 31: For this I can we can use `console.warn` or `console.error` for better view in the Dev console

nextjs-13/src/repositories/newsletter.repository.test.ts
```diff
@@ -0,0 +1,24 @@
+import { dbClientMock } from '@test/database';
+import { newsletterFactory } from '@test/factories/newsletter.factory';
+
+import { createNewsletter } from './newsletter.repository';
+
+describe('Newsletter Respository', () => {
+  describe('createNewsletter', () => {
+    it('creates a newsletter', async () => {
+      const newsletterAttributes = {
```
phisith, line 9: Why you define an attribute here, actually just mock the data should be enough right ?

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: Oh I deliberately want to leave this here so that I can view all the bodies of the requests 👀

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: It's because that, for `POST` requests, the `req` will have the body. But for `GET` requests, the requests not have body, so when it executes `req.json()`, it doesn't just return an empty hash `{}`, but it raises error. So this line just means that if we can extract the body, then extract(`POST`). Else(`GET`), don't do anything, and don't raise error too please 😄 .

nextjs-13/src/repositories/newsletter.repository.test.ts
tednguyendev: Yes I don't need the `newsletterAttributes` for the mocking, but I need it for the [`createNewsletter` method](https://github.com/nimblehq/js-framework-benchmark/pull/96/files#diff-7f62d1d6ff486da9e5be846b0de3dc572e91704e2e913d9aa47ab9e57090f37bR4), which only accepts `Newsletter`, means that it only 3 fields `name`, `content`, and `user`.

Ah actually my current `newsletterAttributes` now is not correct, it should have the field `user` instead of `userId`. Let me fix this.

tednguyendev: Fixed: https://github.com/nimblehq/js-framework-benchmark/pull/96/commits

nextjs-13/src/lib/handler/base.handler.ts
phisith:
```
 if (req.method === 'POST') {
    // Process a POST request
  }
```
maybe you can try this instead, not sure it is work in App route [Reference](https://nextjs.org/docs/pages/building-your-application/routing/api-routes).

nextjs-13/src/app/api/v1/newsletter/route.test.ts
tednguyendev: Fixed [here](https://github.com/nimblehq/js-framework-benchmark/pull/96/commits/3a859900a653b0f79eb980740f865e6ddb78960e#diff-2eff60ff63f1485bc115ee04aaa6cba7fd6bf97cdbe7479ae73fbce0cafb6b52L10)

nextjs-13/src/app/api/v1/me/route.test.ts
tednguyendev: Fixed [here](https://github.com/nimblehq/js-framework-benchmark/pull/96/commits/7e83686ee97e415876c573b274caaf60c3984dd6#diff-2ee6b5c4cbf359c8cda60ecacf45024db9a86bbe0eaa6ad5d51dfb7ee9780aa4R13)

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: Improved [here](https://github.com/nimblehq/js-framework-benchmark/pull/96/commits/7e83686ee97e415876c573b274caaf60c3984dd6#diff-add2a84e0fcd856a4db7e6914984152b0d780f178b1988d446d1d0d0a9e7eac6R31)

nextjs-13/src/app/api/v1/newsletter/route.test.ts
tednguyendev: Oh actually maybe not 😄 
CleanShot 2023-06-01 at 16 06 03

phisith: Oh Ok, sorry that my bad 😅

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: Ye I think so. In ruby, the equivalent of `console.log` is `print`, but when people want to log things in production, they use [`Logger`](https://ruby-doc.org/stdlib-2.4.0/libdoc/logger/rdoc/Logger.html) instead. So I think maybe we only use `console.log` at the local 👀

phisith: Noted, Thanks for the knowledge 😆

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: Yeah I have thought of that too, but if we want to use if else statements instead, we then also have to use if else for `PUT`, `PATCH`, `DELETE`, ... The try-catch seems lazy but I think it may help us make the code less verbose a bit, no? 👀

phisith: Ya in this case try-catch make the code less verbose, maybe just use try-catch should be fine here.

nextjs-13/prisma/schema.prisma
```diff
@@ -8,13 +8,28 @@ datasource db {
...
+}
+
+model Newsletter {
+  id        String   @id @default(uuid())
+  name      String   @db.VarChar(255)
+  content   String   @db.Text()
+  user      User     @relation(fields: [userId], references: [id])
+  userId    String   @map("user_id")
+  createdAt DateTime @default(now()) @map("created_at")
+  updatedAt DateTime @default(now()) @map("updated_at")
```
phisith, line 31: For the `@updateAt` I think you can the field type to @updateAt this auto update when the row got an update. [Reference](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat)

tednguyendev: Ah right, I forgot that. Good catch, thank you! 🚀 

The fix: https://github.com/nimblehq/js-framework-benchmark/pull/96/commits/9f9d399a0e891cb765de1e5166e04419518888c7

phisith: Maybe on the user model too

phisith: Oh you already added, Sorry I couldn't see in the fix file 😅

nextjs-13/src/lib/handler/base.handler.ts
olivierobert: This should be removed. This is fine during development but for production load, we normally https://github.com/winstonjs/winston or https://github.com/pinojs/pino

nextjs-13/src/repositories/newsletter.repository.ts
```diff
@@ -0,0 +1,9 @@
+import { Prisma } from '@prisma/client';
+
+import dbClient from '../config/database';
+
+const createNewsletter = async (
+  newsletterAttributes: Prisma.NewsletterCreateInput
```
tednguyendev, line 6: ^One improvement thanks to @phisith's suggestion 😄

nextjs-13/src/repositories/newsletter.repository.test.ts
```diff
@@ -0,0 +1,24 @@
...
+        name: 'iOS 17 release',
+        content:
+          'iOS 17 will reportedly have a new feature that turns locked iPhones into smart home displays. The interface will display information such as appointments, the weather, and notifications. It will operate similarly to smart home devices from other companies.',
+        user: { connect: { id: '1' } },
```
olivierobert, line 13: Is `connect` required by Prisma? I am unsure how this works for association 🤔

nextjs-13/src/lib/handler/base.handler.ts
```diff
@@ -0,0 +1,37 @@
+import { User } from '@prisma/client';
+import { NextResponse, NextRequest } from 'next/server';
+import { getToken } from 'next-auth/jwt';
+
+import { findUserById } from 'repositories/user.repository';
+
+export default async function baseHandler(
```
olivierobert, line 7: `base` would mean that it could be extended (with more features). How about naming it `appHandler` or `defaultHandler` instead? In addition, since we are checking for user authentication, it could be named `authRequiredHandler`.

nextjs-13/src/lib/handler/base.handler.ts
```diff
@@ -0,0 +1,37 @@
...
+        return await callback(currentUser, body);
+      }
+    }
+
+    return NextResponse.json({ message: 'Invalid token' }, { status: 401 });
+  } catch (err) {
+    console.error('========>err : ', err);
+    return NextResponse.json(
+      { message: (err as Error).message },
+      { status: 500 }
```
olivierobert, line 34: `500` errors should never be raised by the application on purpose. `500` is reserved for unhandled errors. Here we could raise a `400` error (bas request) 💡

nextjs-13/src/lib/handler/base.handler.ts
olivierobert: I am still. not convinced 😅 Prefer early returns over long if/else blocks + rely on the catch block to handle callback errors:

```javascript
const token = await getToken({ req });
const currentUser = await findUserById(token?.userId as string);

if (!currentUser) {
  // raise auth error
}

return await callback(currentUser, req.json());
```

olivierobert (changes requested):

nextjs-13/src/repositories/newsletter.repository.test.ts
tednguyendev: Yes it is, here are [some examples](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#examples-25).

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: I see. Yes, these would be better names, let me update 👀

nextjs-13/src/app/api/v1/newsletter/route.ts
```diff
@@ -0,0 +1,23 @@
...
+      const attributes = {
+        name: name,
+        content: content,
+        user: { connect: { id: currentUser.id } },
+      };
+      const record = await createNewsletter(attributes);
+
+      return NextResponse.json({ newsletter: record }, { status: 200 });
+    } catch (err) {
+      return NextResponse.json({ message: 'Invalid params' }, { status: 422 });
```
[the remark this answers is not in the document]
tednguyendev: I think it doesn't have that(or I can not find it). As I checked, the current type of error is the [PrismaClientValidationError ](https://www.prisma.io/docs/reference/api-reference/error-reference#prismaclientvalidationerror), which only support the `message` method. And when I try to use it, the output will be something like this 😭 

CleanShot 2023-06-02 at 14 41 48@2x

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: But I really want to use this code to handle the unhandled errors 🤔 If we use 400, and the errors is on our side, then I think it may cause confusion to the client. What do you think about this? Or should we use other `5xx` error status codes?

And yeah, seems like I forgot to use the `http-status-codes` here, let me update it :D

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: Yeah I think early returns are better too.

So to keep this simple, I will let the APIs handle the `req.json()` part themself for now, so now the code is simplified into [this](https://github.com/nimblehq/js-framework-benchmark/pull/96/files#diff-10b91e667909ffe6c8b2d34d735e0fa46be166ba37cf1b5ec24f7ea6b54658bbR12-R22).

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: The removed commit: https://github.com/nimblehq/js-framework-benchmark/pull/96/commits/455c612b291d7e0f3ddf58b4f2b24ae0f3e08756#diff-10b91e667909ffe6c8b2d34d735e0fa46be166ba37cf1b5ec24f7ea6b54658bbL24

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: Done! [And this is the new commit](https://github.com/nimblehq/js-framework-benchmark/pull/96/commits/455c612b291d7e0f3ddf58b4f2b24ae0f3e08756#diff-10b91e667909ffe6c8b2d34d735e0fa46be166ba37cf1b5ec24f7ea6b54658bbR8)

nextjs-13/src/lib/handler/base.handler.ts
tednguyendev: [^ Done using http status codes instead ✅ ](https://github.com/nimblehq/js-framework-benchmark/pull/96/commits/455c612b291d7e0f3ddf58b4f2b24ae0f3e08756#diff-10b91e667909ffe6c8b2d34d735e0fa46be166ba37cf1b5ec24f7ea6b54658bbR27)

nextjs-13/src/app/api/v1/newsletter/route.ts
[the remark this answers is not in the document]
olivierobert: Indeed, it is [an open issue on Prisma](https://github.com/prisma/prisma/issues/5040#issuecomment-1343330806). Another framework (Redwood JS) created a wrapper on top of Prisma to fix this use: https://redwoodjs.com/docs/redwoodrecord 

Maybe we could give q quick try with: https://www.npmjs.com/package/prisma-better-errors. There is no source code so I am unsure about the quality of the implementation. If not, we can revisit this issue later. It is not a blocker.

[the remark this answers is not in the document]
tednguyendev: Oke let me try it 👀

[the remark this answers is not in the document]
tednguyendev: I just added some new console logs to try to see some of it's format:

image

But seems like the result are not very good either 😨 :

image

Seems like they are more suitable for other type of [Prisma errors](https://www.prisma.io/docs/reference/api-reference/error-reference#prismaclientvalidationerror), not our [PrismaClientValidationError](https://www.prisma.io/docs/reference/api-reference/error-reference#prismaclientvalidationerror) 🤔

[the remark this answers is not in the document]
olivierobert: Alright, then let's keep it like this for now

nextjs-13/src/lib/handler/app.handler.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    if (!currentUser)
+      return NextResponse.json(
+        { message: 'Invalid token' },
+        { status: StatusCodes.UNAUTHORIZED }
+      );
+
+    return await callback(currentUser, req);
+  } catch (err) {
+    console.error('========>err : ', err);
```
olivierobert, line 24: So you still want to keep this? 😅 Can't we at least only have it in the development environment?

tednguyendev: Ah no let me remove it, I just miss it 😄

olivierobert (approved):

phisith (approved):

PR #104, opened 2023-05-31
[#40][#41] [Backend][Frontend] As a user I can logout

tednguyendev:
Close both #40 and #41 since not need to change the backend

## What happened 👀
- Add a dropdown component to display a logout button.
- Add a new icon component to render SVG icon images.
- Trigger `NextAuth`'s `signOut` function when click the logout button.

## Insight 📝
Since we only need to use the `NextAuth`'s `signOut` function to log out, we really don't need to change anything in the backend.

### Workflow
https://github.com/nimblehq/js-framework-benchmark/assets/76901736/4b73acee-8ceb-44bb-a87e-02984e2c236f

### Cookies
Before sign out:

CleanShot 2023-05-29 at 12 18 34@2x

After sign out:

CleanShot 2023-05-29 at 12 19 03@2x


phisith (approved):
Clean work for me 👌🏼

nextjs-13/src/components/Dropdown/index.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+      render(<Dropdown />);
+
+      fireEvent.click(screen.getByTestId('dropdown-toggle'));
+
+      await waitFor(() => {
+        expect(screen.getByTestId('dropdown-menu')).toBeTruthy();
+      });
+
+      expect(screen.getByText('Sign out')).toBeInTheDocument();
+    });
```
olivierobert, line 21: Let's add a line break for readability (you seemed to hate line breaks 🙈 )

olivierobert (approved):
Just a minor code style issue but pre-approving

nextjs-13/src/components/Dropdown/index.test.tsx
tednguyendev: Oops 🙈 Done! https://github.com/nimblehq/js-framework-benchmark/pull/104/commits/f7da8bf4902c6ba26297d0388ac4912b860cf013. Eslint not auto add line break when format files so I often forget too haha.

olivierobert (approved):

PR #105, opened 2023-05-31
[#59] [Backend] As a user I can logout

phisith:
Close [#59]

## What happened 👀

- Implement Logout with Testing

## Insight 📝

N/A

## Proof Of Work 📹
Screenshot 2566-05-31 at 12 11 44

### DIFF
--- a/.github/workflows/remix_test.yml
+++ b/.github/workflows/remix_test.yml
@@ -29,6 +29,19 @@ jobs:
           cache: "npm"
           cache-dependency-path: "remix/package-lock.json"
 
+      - name: Create env file
+        run: |
+          touch .env
+          echo NODE_ENV=development >> .env
+          echo GOOGLE_CLIENT_ID=274093098276-7sar8nsfcpsge9ni7p4nbi5rui6ed3dn.apps.googleusercontent.com >> .env
+          echo GOOGLE_CLIENT_SECRET=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 >> .env
+          echo SESSION_NAME=__session >> .env
+          echo SESSION_SAMESITE=lax >> .env
+          echo SESSION_PATH=/ >> .env
+          echo SESSION_HTTPONLY=1 >> .env
+          echo SESSION_SECRETS=veryscret >> .env
+          echo SESSION_SECURE=1 >> .env
+
       - name: Install dependencies
         run: npm ci
 
--- a/.github/workflows/remix_test.yml
+++ b/.github/workflows/remix_test.yml
@@ -29,19 +29,6 @@ jobs:
           cache: "npm"
           cache-dependency-path: "remix/package-lock.json"
 
-      - name: Create env file
-        run: |
-          touch .env
-          echo NODE_ENV=development >> .env
-          echo GOOGLE_CLIENT_ID=274093098276-7sar8nsfcpsge9ni7p4nbi5rui6ed3dn.apps.googleusercontent.com >> .env
-          echo GOOGLE_CLIENT_SECRET=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 >> .env
-          echo SESSION_NAME=__session >> .env
-          echo SESSION_SAMESITE=lax >> .env
-          echo SESSION_PATH=/ >> .env
-          echo SESSION_HTTPONLY=1 >> .env
-          echo SESSION_SECRETS=veryscret >> .env
-          echo SESSION_SECURE=1 >> .env
-
       - name: Install dependencies
         run: npm ci
 
--- a/remix/app/routes/auth.sign-out.test.tsx
+++ b/remix/app/routes/auth.sign-out.test.tsx
@@ -9,6 +9,7 @@ import { Authenticator } from 'remix-auth';
 
 import { action } from './auth.sign-out';
 import { sessionFactory } from '../tests/factories/session.factory';
+import { makeRequest } from '../tests/helpers/request';
 
 jest.mock('./auth.sign-out', () => ({
   action: jest.fn(),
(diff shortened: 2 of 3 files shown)


.github/workflows/remix_test.yml
```diff
@@ -29,6 +29,19 @@ jobs:
           cache: "npm"
           cache-dependency-path: "remix/package-lock.json"
 
+      - name: Create env file
+        run: |
+          touch .env
+          echo NODE_ENV=development >> .env
+          echo GOOGLE_CLIENT_ID=274093098276-7sar8nsfcpsge9ni7p4nbi5rui6ed3dn.apps.googleusercontent.com >> .env
```
tednguyendev, line 36: I think this is quite dangerous as we are exposing our credentials through our code base 👀, especially while our repo is also a public repo too.

remix/.env.sample
```diff
@@ -1,7 +1,6 @@
 DATABASE_URL="postgresql://postgres:postgres@localhost:5432/mydb?schema=public"
 NODE_ENV=development
-GOOGLE_CLIENT_ID=XXXXXX
-GOOGLE_CLIENT_SECRET=XXXXXX
+GOOGLE_CLIENT_ID=274093098276-7sar8nsfcpsge9ni7p4nbi5rui6ed3dn.apps.googleusercontent.com
 GOOGLE_CLIENT_SECRET=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
```
tednguyendev, line 4: Same as above 🤔

remix/app/routes/auth.sign-out.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { createCookieSessionStorage } from '@remix-run/node';
+import { Authenticator } from 'remix-auth';
+
+import { action } from './auth.sign-out';
+
+describe('Auth sign-out', () => {
+  const sessionStorage = createCookieSessionStorage({
+    cookie: { secrets: ['s3cr3t'] },
```
tednguyendev, line 14: Shouldn't the 's3cr3t' be in .env instead? 👀

remix/.env.sample
phisith: For this one I can remove it

Solved on: [5aace66](https://github.com/nimblehq/js-framework-benchmark/pull/105/commits/5aace6609fb6189d6550794b013ac455a01287c9)

remix/app/routes/auth.sign-out.test.tsx
phisith: For this one and the env in remix_test.yml, For the reason doing this is because my testing on Github CI can't read the .env file or env variable.

To solve this I might need to get the admin permission on the repo and create the create variable [Reference](https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository), or is there any other way.

tednguyendev: Which are the tests that need .env variable?

Can you mock these tests so that they do not need the `.env`?  👀 I think mocking will be a better practice in this case.

phisith: In this case, you can see the `action` function in the `routes/auth.sign-out.tsx` and it use a authenticator, so that authenticator need .env to work.

So here I should just mock the `action` right ?

tednguyendev: I see.

I only unit test in the Next.js v13 application instead of integration test like you, but for this, I remember Olivier mocked the config file [like this](https://github.com/nimblehq/js-framework-benchmark/blob/develop/nextjs-12/src/lib/middleware/passport.middleware.ts#L49) in the Next.js v12 application.

Maybe you can take a look at this and then you can ask for more advices from @olivierobert 😄. I don't have much experience with testing on the Frontend too 🥲

phisith: Ok 😅, I will take a look at it. Thank you for the suggestion 🙏🏼.

phisith: Hi @tednguyendev, I think for now I just do it this way [ae0d6f5](https://github.com/nimblehq/js-framework-benchmark/pull/105/commits/ae0d6f5d6e7347cdc2e4a0c13fff8feaf492a24f). Not sure is it correct or fine but just to avoid showing the env stuff in github CI file 🙉.

tednguyendev: Oke! LGTM! 😆

remix/app/routes/auth.sign-out.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  (action as jest.Mock).mockReturnValue({
+    headers: {
+      location: '/auth/sign-in',
+      'Set-Cookie': sessionMock.session,
+    },
+  });
+
+  describe('POST /auth/sign-out', () => {
+    it('returns to /auth/sign-in', async () => {
+      const request = new Request('http://localhost:3400/auth/sign-out', {
```
olivierobert, line 33: I was initially surprised why we need to pass an absolute URL here but then I found https://github.com/remix-run/remix/discussions/993#discussioncomment-4125164 (this should probably go into the insight section of the pull request for others to understand this "strange" bit 💡)

To avoid repetition and reduce maintenance issues in the long run, I recommend creating a test helper in `tests/helpers/request.ts` so that we can use something like `makeRequest('/auth/sign-out', 'POST') in tests.

olivierobert (changes requested):

remix/app/routes/auth.sign-out.test.tsx
phisith: Yes, for the first I also got the error from that, then I just passed the absolute URL there.

Solved on: [d094998](https://github.com/nimblehq/js-framework-benchmark/pull/105/commits/d0949985162ee27c4eb6523b2f893a17a4d8b2bb)

tednguyendev (approved):

olivierobert (approved):

olivierobert:
@phisith rebase required for this to be merged

phisith:
> @phisith rebase required for this to be merged

Resolved.

PR #106, opened 2023-05-31
[#45] [Frontend] As a user I can create a newsletter

tednguyendev:
#45 

## What happened 👀
- Add layout for Newsletter CRUD
- Add button to open create Newsletter modal
- Add create Newsletter modal
- Add tests

## Insight 📝
To keep it simple, I mainly create components and style manually, plus using some libraries to add some animations like `react-modal` + `react-spinners` + `react-toastify` and`react-textarea-autosize` to easily auto-grow the textarea that contains the content.

## Proof Of Work 📹
https://github.com/nimblehq/js-framework-benchmark/assets/76901736/cc586946-e1c1-4aa9-9573-047843243d1a


nextjs-13/src/app/page.tsx
```diff
@@ -15,12 +20,29 @@ const Home = () => {
...
       <Head>
         <title>Welcome to NextNewsletter 🚀</title>
       </Head>
-      <h4>Welcome to NextNewsletter 🚀</h4>
-    </>
+      <div className="home__nav">
+        <div className="home__tab">Newsletter</div>
+      </div>
+      <div className="home__dashboard">
+        <div></div>
```
phisith, line 31: I think you can remove this empty `div`

nextjs-13/src/components/CreateNewsletterModal/index.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const TextareaAutosizeMock = jest.fn((props) => (
+    <textarea
+      id="content"
+      onChange={(event) => {
+        props.onChange(event); // Call the provided onChange callback
+      }}
+      value={props.value}
+    />
+  ));
+  return TextareaAutosizeMock;
```
phisith, line 32: I think you can add line break above this `return`

nextjs-13/src/components/CreateNewsletterModal/index.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+});
+
+describe('CreateNewsletterModal', () => {
+  it('renders without modal initially', () => {
+    render(
+      <CreateNewsletterModal modalIsOpen={false} setIsOpen={() => undefined} />
+    );
+    const modalElement = screen.queryByTestId('modal-content');
```
phisith, line 40: And add link break between `render` `const` `expect` could make the code more easier to read 👀

nextjs-13/src/components/CreateNewsletterModal/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+      await requestManager('POST', 'v1/newsletter', {
+        data: {
+          name,
+          content,
+        },
+      });
+
+      closeModal();
+      setLoading(false);
+      toast.success('Created newsletter success!', {
```
phisith, line 51: For `toast` I think is better to move it to `/lib/toast/makeToast.ts` to make it more reusable without define the styling or position.

And for the `ToastContainer` you can just wrap it in `app.tsx` or `layoujt` as you would need to call toast in other page too.

So after that you just call `makeToast('message', 'error')` without import the `ToastContainer` in every page.

nextjs-13/src/components/CreateNewsletterModal/index.test.tsx
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/106/commits/b5c4314bf5475947150344015ba6bcdc399813cc#diff-8a0a03315819434e30ec3761666c2fd68166d67d0332a15196e7f69df14e5eb7R32)

nextjs-13/src/app/page.tsx
tednguyendev: Actually I need that `div` there.

It's because I want the [`Create Newsletter`](https://github.com/nimblehq/js-framework-benchmark/pull/106/files#diff-223e9b714cf2c9b5f163079f0bd63d3a6eac9c46915fc3afde20a2742897d706R32-R38) button to be at the utmost right, like this:

CleanShot 2023-06-06 at 13 32 18@2x

And I think that I should use `display: flex` to do this, so I'm using the parent(the one with className `home__dashboard`) [with `display: flex` and `justify-content: space-between`](https://github.com/nimblehq/js-framework-benchmark/pull/106/files#diff-e8ab88801c53a9530e8eca7be803e279c44fa9d639d72b594a0b809a24bcfad4R30), and putting the empty div there so that it be at the utmost left, and the [`Create Newsletter`](https://github.com/nimblehq/js-framework-benchmark/pull/106/files#diff-223e9b714cf2c9b5f163079f0bd63d3a6eac9c46915fc3afde20a2742897d706R32-R38) button at the utmost right.

CleanShot 2023-06-06 at 13 32 18@2x

Anyway, I [changed it from `<div></div>` to `<div />`](https://github.com/nimblehq/js-framework-benchmark/pull/106/commits/b5c4314bf5475947150344015ba6bcdc399813cc#diff-223e9b714cf2c9b5f163079f0bd63d3a6eac9c46915fc3afde20a2742897d706R31) to looks a bit more concise 👀

nextjs-13/src/components/CreateNewsletterModal/index.test.tsx
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/106/commits/b5c4314bf5475947150344015ba6bcdc399813cc#diff-8a0a03315819434e30ec3761666c2fd68166d67d0332a15196e7f69df14e5eb7R51)

nextjs-13/src/components/CreateNewsletterModal/index.tsx
tednguyendev: I see 👀

Doing that indeed make the code looks a lot better! Thank you for this suggestion, Beckham! 🔥

Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/106/commits/c312e7e088f83d969ab2a05cf1e01f8cd58c15af)

nextjs-13/src/app/page.tsx
phisith: Okie understood, that's better

nextjs-13/src/components/CreateNewsletterModal/index.tsx
phisith: Actually this suggestion is come from when I did the code challenge with @olivierobert 😅.

We need to thank you him also 😂.

nextjs-13/src/app/page.tsx
tednguyendev: Wait, TIL that `flex-end` exists 🤣

Improved [here](https://github.com/nimblehq/js-framework-benchmark/pull/106/commits/a41562135f8d8a1b30e2e0e1025831022f64e27a).

phisith (approved):

nextjs-13/src/components/CreateNewsletterModal/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    setName('');
+    setContent('');
+  };
+
+  const handleSubmit = async (event) => {
+    event.preventDefault();
+    setLoading(true);
+
+    try {
+      await requestManager('POST', 'v1/newsletter', {
```
olivierobert, line 41: As raised in #63, it would have been acceptable too to use [Next.JS server action (backend)](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions) so as not to create a Single Page Application (UI communicating over an API) 😅

tednguyendev: I see 👀 So what do you think about this? Should we prefer implementing server actions to UI communicating over API in this POC? 🤔

olivierobert: Yes, I think so as it is one of the major benefit of Next.JS 13. It simplifies the implementation a lot and it looks closer to what we would do in Rails or any full-stack framework. It is also the direction most React frameworks go to. SPA is out, server rendering is in 💯

olivierobert (approved):

nextjs-13/src/components/CreateNewsletterModal/index.tsx
tednguyendev: Understood!

Seems like there're so many trending frontend technologies I need to catch up 🙈 

As I already implemented the SPA way for nearly all the tasks left(they are in draft PRs now), I think for now, I should use the server-side action for the only remaining task, which is the task about viewing a newsletter as a guest.

Then I will create another PR to refactor all SPA code into the server-side action way.

Does that sound good? 👀

phisith: Hi @tednguyendev, if you're going to do the server-side fetching, I would suggest you to try out this library [useSWR](https://swr.vercel.app/). As it implemented well with the server component for the NextJs, and it compatible with `axios` too.

From the example, it help you to manage the loading and error (better than create our own react state).
```
import useSWR from 'swr'
 
function Profile() {
  const { data, error, isLoading } = useSWR('/api/user', fetcher)
 
  if (error) return <div>failed to load</div>
  if (isLoading) return <div>loading...</div>
  return <div>hello {data.name}!</div>
}
```

tednguyendev: Thanks you for the suggestion 👀 I will take a lot at it! 😄

tednguyendev: UPDATED: Working on [this PR](https://github.com/nimblehq/js-framework-benchmark/pull/129) to refactor to the Server Actions way. It's very easy to refactor and it looks SUPER neat 👀

PR #107, opened 2023-06-01
[#42] [Backend] As a user I can list newsletters

tednguyendev:
#42 

## What happened 👀
- Add `queryNewsletterByUserId` to Newsletter Repository
- Add new route to get a list of newsletters
- Add tests

## Insight 📝
N/A

## Proof Of Work 📹
CleanShot 2023-06-01 at 09 49 09@2x

### DIFF
--- a/nextjs-13/src/repositories/newsletter.repository.ts
+++ b/nextjs-13/src/repositories/newsletter.repository.ts
@@ -6,4 +6,13 @@ const createNewsletter = async (
   newsletterAttributes: Prisma.NewsletterCreateInput
 ) => dbClient.newsletter.create({ data: newsletterAttributes });
 
-export { createNewsletter };
+const queryNewsletterByUserId = async (id: string) => {
+  const user = await dbClient.user.findUnique({
+    where: { id: id },
+    include: { newsletters: true },
+  });
+
+  return user?.newsletters;
+};
+
+export { createNewsletter, queryNewsletterByUserId };
--- a/nextjs-13/prisma/schema.prisma
+++ b/nextjs-13/prisma/schema.prisma
@@ -8,13 +8,13 @@ datasource db {
 }
 
 model User {
-  id         String       @id @default(uuid())
-  name       String       @db.VarChar(100)
-  email      String       @db.VarChar(50)
-  avatarUrl  String?      @map("avatar_url")
-  createdAt  DateTime     @default(now()) @map("created_at")
-  updatedAt  DateTime     @default(now()) @map("updated_at") @updatedAt
-  Newsletter Newsletter[]
+  id          String       @id @default(uuid())
+  name        String       @db.VarChar(100)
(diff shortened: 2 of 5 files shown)


nextjs-13/src/repositories/newsletter.repository.ts
```diff
@@ -6,4 +6,13 @@ const createNewsletter = async (
   newsletterAttributes: Prisma.NewsletterCreateInput
 ) => dbClient.newsletter.create({ data: newsletterAttributes });
 
-export { createNewsletter };
+const queryNewsletterByUserId = async (id: string) => {
```
phisith, line 9: I think you can change this function name to `findNewsletterByUserId` instead of `queryNewsletterByUserId`.  This makes it more like natural language and easier to understand.

[Reference](https://stackoverflow.com/questions/2141818/method-names-for-getting-data)

phisith (approved):
Overall LGTM! 👍🏼

nextjs-13/src/repositories/newsletter.repository.ts
tednguyendev: Ya although `query` does not sound as easy as `find`, I think `find` is more suitable for cases when we only need to get only one record, and `query` is something that is more suitable for getting many records at once, like, when we want to get some data(joins table bla bla) by doing some raw sql, we usually says "query sql".

And maybe the reason I think so is because I'm so used to Ruby on Rails, and RoR use [`find` to only get one record](https://apidock.com/rails/ActiveRecord/Base/find/class), and [`where` to get many records](https://apidock.com/rails/ActiveRecord/QueryMethods/where) 😄. I'm an advocate for using 2 different verbs for 2 different kinds of getting data(get one and many), as it helps us need less brain power.

For example, in Rails way, we just need to read the first verb of a function name (`find` or `query`) to know if it queries 1 record or many. If we use the same `find` for both kinds of query, like `findNewsletter` and `findNewsletters`, we have to read until the last `s` word to know that if this is getting one or many 🤔. `findNewsletter` and `queryNewsletters` are faster to differentiate to me.

So this is just a personal preference thing, I think 👀

phisith: Oh okay, Thanks for sharing 😁. 

But for this function, you are using `findUnique` in Prisma, which will return only one record, so I think `find` will be good here as you mention the use case to use `find` above.

tednguyendev: Ya, but the thing this function return is the newsletters of the user, which is plural 👀 It may cause confusion a bit here, as in later PRs, I refactored it to be:

```typescript
dbClient.newsletter.findMany({
    where: { userId: userId }
})
```

phisith: Noted

olivierobert (approved):

PR #110, opened 2023-06-01
[#43] [Frontend] As a user I can list newsletters

tednguyendev:
#43

## What happened 👀
- Add `ListNewsletter` component + style + connect API
- Use `Suspense` to loads `ListNewsletter`
- Add tests

### React's [`Suspense`](https://react.dev/reference/react/Suspense)
At first, I tried to use the `Suspense` to see if it can simplify our code a bit(not need to have states like `isLoading`). But then I figurated out that it's not very suitable for our usecase, and to keep this POC simple as discussed in the comments, I removed it.

The following are my notes on how I used `Suspense`. If needed, you can view the code of [this last commit before I removed the `Suspense`](https://github.com/nimblehq/js-framework-benchmark/tree/ffdcdec00aeb0823d17caa4c8d5f225cd55d97ed).

To use `Suspense`:

First, [this `promiseWrapper`](https://github.com/nimblehq/js-framework-benchmark/blob/ffdcdec00aeb0823d17caa4c8d5f225cd55d97ed/nextjs-13/src/app/page.tsx#L26) is exclusively to make `Suspense` work.

When the component inside `Suspense` is fetching data (`ListNewsletter`), we need to [throw a pending `Promise`](https://github.com/nimblehq/js-framework-benchmark/blob/ffdcdec00aeb0823d17caa4c8d5f225cd55d97ed/nextjs-13/src/lib/request/promiseWrapper.tsx#L20) so that `Suspense` can catch and can trigger the loading. That's the purpose of the `promiseWrapper` also.

And as I tested, it's true that without the `promiseWrapper`, the `Suspense` is never triggered and the `Loading` never appears.

As I research about `Suspense`, I mainly use these 2 blog posts:
- [What Is React Suspense](https://prismic.io/blog/what-is-react-suspense)
- [React Suspense for Data Fetching with Axios](https://deadsimplechat.com/blog/react-suspense/)

## Proof Of Work 📹

https://github.com/nimblehq/js-framework-benchmark/assets/76901736/b85cb4ec-878f-4ba3-8a1b-c07101fe906d


phisith (approved):
LGTM !!

nextjs-13/src/app/page.tsx
```diff
@@ -28,6 +44,12 @@ const Home = () => {
         <div className="home__tab">Newsletter</div>
       </div>
       <div className="home__dashboard">
+        <div>
+          <h3>Your Newsletters</h3>
+          <Suspense fallback={<ClipLoader loading={true} size={150} />}>
```
olivierobert, line 49: That loader is quite massive 🤣 We could probably decrease its size a bit

nextjs-13/src/lib/request/promiseWrapper.tsx
```diff
@@ -0,0 +1,32 @@
+const promiseWrapper = (promise) => {
+  let status = 'pending';
+  let result;
+
+  const s = promise.then(
```
olivierobert, line 5: Also spell out the complete variable name. Avoid cryptic variable naming: https://nimblehq.co/compass/development/code-conventions/#convention-12

nextjs-13/src/stylesheets/components/_list-newsletter.scss
```diff
@@ -0,0 +1,23 @@
+.list-newsletter {
+  font-weight: var(--font-weight-5);
+  width: 500px;
+
+  ul {
```
olivierobert, line 5: Avoid targetting HTML tags directly, prefer using classes (these are less strong precedence)

nextjs-13/src/stylesheets/components/_list-newsletter.scss
```diff
@@ -0,0 +1,23 @@
...
+      display: flex;
+      flex-direction: row;
+      background: var(--gray-0);
+      border-radius: var(--radius-conditional-2);
+      padding: var(--size-2);
+      color: black;
+      cursor: pointer;
+      --_size: var(--font-size-1);
+      margin-bottom: var(--size-3);
+      box-shadow: var(--shadow-2);
```
olivierobert, line 20: There is room to improve the ordering of the styles for readability:

```css
    li {
     // always place variables at the top of the block
      --_size: var(--font-size-1); 

      // place any style that can affect the positioning first
      display: flex;
      flex-direction: row;
      margin-bottom: var(--size-3);
      padding: var(--size-2);

      // font related styles nexts
      color: black;
      
      // the rest
      cursor: pointer;
      background: var(--gray-0);
      border-radius: var(--radius-conditional-2);
      box-shadow: var(--shadow-2);
   }

```

nextjs-13/src/components/ListNewsletter/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface PromiseWrapper {
+  read: () => Record[];
+}
+
+interface Props {
+  promise: PromiseWrapper;
+}
+
+const ListNewsletter = ({ promise }: Props) => {
```
olivierobert, line 14: Why pass a promise? That is kind of unexpected as a prop. The most standard way would be to pass an array of newsletters. It adds complexity for testing the component (https://github.com/nimblehq/js-framework-benchmark/pull/110/files#diff-fbbd0021aabad22d799d758d4fc16f2c26225f683f50b56c7cbbb068cbb1c774R22) for no apparent benefits

nextjs-13/src/app/page.tsx
```diff
@@ -1,20 +1,36 @@
...
-  if (status === 'loading') {
-    return <p>Loading...</p>;
+  async function fetchRecords() {
+    const res = await requestManager('GET', 'v1/newsletter');
+
+    return res.records;
+  }
+
+  async function getData() {
+    setPromise(promiseWrapper(fetchRecords()));
```
olivierobert, line 26: I do not understand why this `promiseWrapper` is required?

nextjs-13/src/app/page.tsx
```diff
@@ -1,20 +1,36 @@
...
 
 const Home = () => {
   const { status } = useSession();
   const [modalIsOpen, setIsOpen] = useState(false);
+  const [promise, setPromise] = useState();
 
-  if (status === 'loading') {
-    return <p>Loading...</p>;
+  async function fetchRecords() {
+    const res = await requestManager('GET', 'v1/newsletter');
```
olivierobert, line 20: Ditto from below. Avoid abbreviations: https://nimblehq.co/compass/development/code-conventions/#convention-12

```suggestion
    const newsletterList = await requestManager('GET', 'v1/newsletter');
```

nextjs-13/src/stylesheets/components/_list-newsletter.scss
```diff
@@ -0,0 +1,23 @@
+.list-newsletter {
+  font-weight: var(--font-weight-5);
+  width: 500px;
```
olivierobert, line 3: Always prefer to use `rem` for unit sizing

nextjs-13/src/app/page.test.tsx
```diff
@@ -1,51 +1,107 @@
...
 
-      expect(screen.getByText('Create newsletter')).toBeInTheDocument();
+      await waitFor(() => {
+        expect(screen.getByText('Your Newsletters')).toBeInTheDocument();
+      });
     });
 
-    it('renders a modal to create newsletter when click create newsletter button', async () => {
-      useSession.mockReturnValue({ status: 'authenticated' });
+    describe('Suspense', () => {
```
olivierobert, line 58: That's a detail of implementation that should not be tests (i.e., it could be removed). In BDD style, focus on describing the behaviors:

```suggestion
    describe('Listing newsletters ', () => {
```

nextjs-13/src/app/page.test.tsx
```diff
@@ -1,51 +1,107 @@
...
+          requestManager.mockImplementation(() => new Promise(() => []));
+        });
+
+        it('renders ClipLoader', async () => {
+          render(<Home />);
+
+          expect(screen.getByTestId('clip-loader')).toBeInTheDocument();
+        });
 
+        it('NOT renders ListNewsletter', async () => {
```
olivierobert, line 70: English syntax 🙈 

```suggestion
        it('does NOT render the list of newsletter', async () => {
```

Again `ListNewsletter` is a detail of implementation that could change (i.e., the component could be renamed) and we don't want to have to update the tests for it.

olivierobert (changes requested):
The insight section of the pull request is empty, but there are many implementation details that could have benefited from being better explained. I have raised several comments because I am unsure why the implementation is done this way.

tednguyendev:
> The insight section of the pull request is empty, but there are many implementation details that could have benefited from being better explained. I have raised several comments because I am unsure why the implementation is done this way.

Ok, let me add the explainations!

It's because this PR is about the frontend and everything is new to me 😄. That's why I wasn't so sure what's new or insightful, as I assumed these things are obvious to others 👀

nextjs-13/src/stylesheets/components/_list-newsletter.scss
tednguyendev: Thank you a lot for these suggestions about the CSS! To be honest, I don't help much experience with the frontend(so the CSS too), so please keep assisting me on my CSS improvement journey 🤣

nextjs-13/src/app/page.tsx
tednguyendev: Ya I should have put the explanation for this on the description.

The `promiseWrapper` is exclusively to make the `Suspense` work.

When the component inside `Suspense` is fetching data(`ListNewsletter`), we need to throw a pending `Promise` so that `Suspense` can catch and can trigger the loading. That's the purpose of the `promiseWrapper` also.

And as I tested, it's true that without the `promiseWrapper`, the `Suspense` is never triggered and the `Loading` never appears.

As I research about `Suspense`, I mainly use this 2 blog posts:
- [What Is React Suspense](https://prismic.io/blog/what-is-react-suspense)
- [React Suspense for Data Fetching with Axios](https://deadsimplechat.com/blog/react-suspense/)

I also read other posts, and all these blog posts say kind of the same thing too!

As for the the reason I pass the `promise` instead of `records` in [this comment](https://github.com/nimblehq/js-framework-benchmark/pull/110#discussion_r1227485608), it's because if I instead pass the `records` to `ListNewsletter`, the `Suspense` never triggers.

Maybe it's because for the `Suspense` to work, the `Promise` need to be thrown INSIDE the component that `Suspense` is wrapping, aka `ListNewsletter`.

I kind of find using `Suspense` a little more troublesome than the normal way(use a `loading` state instead), but indeed it makes our code more concise.

What do you think about this? 👀

olivierobert: This info helps a lot. It does look more complicated, but I also find that the implementation in https://deadsimplechat.com/blog/react-suspense/ (using a hook) is better than the current one in this PR. I am fine with keeping Suspense but we should follow the hook usage instead of passing a promise as a prop (that looks like a code smell in React). As the same time, if you prefer to get rid of the whole Suspense bit, I am good with it too. The choice is yours.

tednguyendev: Actually I tried that way first!

It worked, but it is only applicable if we use that hook in the child component(`ListNewsletter`). I need to use in the parent component(`Home`), because I will need to pass the `getData` function to other components on the incoming PRs to re-render the newsletter list after the user update or delete a newsletter.

Seems like we should go back to the traditional way 👀. I also find `Suspense` seems more suitable for cases when we want to load only an independent component, like, ViewNewsletterDetail, instead of a big list of Newsletters. 😄

nextjs-13/src/app/page.tsx
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/110/commits/d30f9f526aeaccaa44ac7bfcc866291c1952f753#diff-223e9b714cf2c9b5f163079f0bd63d3a6eac9c46915fc3afde20a2742897d706R49).

nextjs-13/src/lib/request/promiseWrapper.tsx
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/110/commits/d30f9f526aeaccaa44ac7bfcc866291c1952f753#diff-f289ac85315a1af5de59576a60350c945a27574bc29040aa1f5607a01b0877fdR5)

nextjs-13/src/stylesheets/components/_list-newsletter.scss
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/110/commits/fe769187ea84c37f4b04d3ba993450abd5940431#diff-099e542ab2be0b36ba6ee7666ef7e5ef285c13f7a6e3e70bb3c854526d41b0e1R3)

nextjs-13/src/stylesheets/components/_list-newsletter.scss
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/110/commits/982cdae493d9ee69a7091072c4dc136019a67cd7#diff-099e542ab2be0b36ba6ee7666ef7e5ef285c13f7a6e3e70bb3c854526d41b0e1R11)

nextjs-13/src/app/page.tsx
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/110/commits/d30f9f526aeaccaa44ac7bfcc866291c1952f753#diff-223e9b714cf2c9b5f163079f0bd63d3a6eac9c46915fc3afde20a2742897d706R20)

nextjs-13/src/stylesheets/components/_list-newsletter.scss
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/110/commits/fe769187ea84c37f4b04d3ba993450abd5940431#diff-099e542ab2be0b36ba6ee7666ef7e5ef285c13f7a6e3e70bb3c854526d41b0e1R3)

nextjs-13/src/app/page.test.tsx
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/110/commits/d30f9f526aeaccaa44ac7bfcc866291c1952f753#diff-4b9ae30a4e34a98e0e0d12af3e4872e7ac9865c68f016c1e6667d4336ce08ce1R70)

nextjs-13/src/app/page.test.tsx
olivierobert: @tednguyendev this comment has yet to be addressed

olivierobert (changes requested):
For the sake of this POC, let's keep the implementation simpler by not using `<Suspense />`. You can extract what you have learned through this project and share it with the JS Guild or in the Web Chapter retrospective. It has value but it is out of scope for this small project.

tednguyendev:
> For the sake of this POC, let's keep the implementation simpler by not using `<Suspense />`. You can extract what you have learned through this project and share it with the JS Guild or in the Web Chapter retrospective. It has value but it is out of scope for this small project.

Oke I got it! At least we now understand more about how `Suspense` works and its use cases 😄

nextjs-13/src/app/page.test.tsx
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/110/commits/da187feeeca914879b6d502720a739b9205f3408#diff-4b9ae30a4e34a98e0e0d12af3e4872e7ac9865c68f016c1e6667d4336ce08ce1R58)

nextjs-13/src/app/page.tsx
tednguyendev: The `Suspense` is removed! Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/110/commits/da187feeeca914879b6d502720a739b9205f3408#diff-223e9b714cf2c9b5f163079f0bd63d3a6eac9c46915fc3afde20a2742897d706R54)

PR #116, opened 2023-06-05
[#49] [Frontend] As a user I can delete a newsletter

tednguyendev:
#49

## What happened 👀
- Separate newsletter items from [`ListNewsletter`](https://github.com/nimblehq/js-framework-benchmark/pull/116/files#diff-4cfa8ba8c408a69f6934a108f1c30bb053cfab26e5f21238c1fa32c887c4a83cL24) into [`NewsletterItem`](https://github.com/nimblehq/js-framework-benchmark/pull/116/files#diff-c6e115565054c1cb22f062088bb3cfc133eddc4cc46b738c9feb26a8785eca35R1)
- Add [delete button into `NewsletterItem`](https://github.com/nimblehq/js-framework-benchmark/pull/116/files#diff-c6e115565054c1cb22f062088bb3cfc133eddc4cc46b738c9feb26a8785eca35R38)
- Add [`sweetalert2`](https://github.com/nimblehq/js-framework-benchmark/pull/116/files#diff-8007718c51aeb6213bc4ff841d1be962ca0462214297fc2eae957e339406444cR45) to use [it's alerts](https://github.com/nimblehq/js-framework-benchmark/pull/116/files#diff-c6e115565054c1cb22f062088bb3cfc133eddc4cc46b738c9feb26a8785eca35R11).
- Add [tests](https://github.com/nimblehq/js-framework-benchmark/pull/116/files#diff-e826a2663ade7bea56375a376ed386fe36acc8da5404a885095a6a8fc0af374fR1) for `NewsletterItem`

## Insight 📝
`N/A`

## Proof Of Work 📹
https://github.com/nimblehq/js-framework-benchmark/assets/76901736/f05867bb-8316-459a-a78c-280103ec4753


nextjs-13/src/components/NewsletterItem/index.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  it('renders the component', async () => {
+    const item = { id: 1, name: 'Newsletter 1' };
+    render(<NewsletterItem item={item} onAfterCloseCallback={null} />);
+
+    expect(screen.getByTestId('newsletter-item')).toBeInTheDocument();
+    expect(screen.getByText(item.name)).toBeInTheDocument();
+    expect(screen.getByTestId('btn-delete')).toBeInTheDocument();
+  });
+
+  it('deletes record and refresh data on delete button click', async () => {
```
phisith, line 25: Not sure, is it better to split this test into 2 case, one for `delete record` and `refresh data` 🤔

```
describe('on delete newsletters')
      it ('delete a newsletter')
      it ('refresh data') # or `call the callback`
```

phisith (commented):
Overall looking great to me, but not sure that my suggestions is help 😅

nextjs-13/src/app/page.tsx
```diff
@@ -17,21 +17,21 @@ const Home = () => {
   const [records, setRecords] = useState([]);
   const [isLoading, setIsLoading] = useState(false);
 
-  useEffect(() => {
-    const getData = async () => {
-      setIsLoading(true);
+  const getData = async () => {
```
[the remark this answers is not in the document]
tednguyendev: I agree that's a better name!

Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/116/commits/5dc46b6f43ed7f2b7361d55ec26393c6dccaa741)

nextjs-13/src/components/NewsletterItem/index.test.tsx
tednguyendev: I was not sure about it too, then I tried to split it into 2 cases to see, and I think it indeed looks better 🙈 

Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/116/commits/d3c440d36eaac7b7ec84c3101f5eef14ed62fcf4)

nextjs-13/src/components/CreateNewsletterModal/index.test.tsx
```diff
@@ -69,7 +81,11 @@ describe('CreateNewsletterModal', () => {
 
   it('enters name', () => {
```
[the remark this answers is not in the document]
phisith: Reference: 
https://www.betterspecs.org/#let
https://www.betterspecs.org/#should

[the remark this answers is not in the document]
tednguyendev: I understand your first comment on this issue, but don't, with the second one 🤣

Jest not having anything like `let`, right? 👀 

And about the `should`, do you means that we should rename the `it('enters name')` into something better? I agree, since `enters name` seems somewhat confusing.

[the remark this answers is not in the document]
phisith: for this first I mean the `it('enters name')` and `it('enters content')` are doing the same things, but now I just realize that you have an input element on `enters name` and textarea element on `enters contents` both are a different element, then the way you have done it should good. 

and Yes, you should rename the `it ('enters name')` to something better like `sets the name field`

Sorry, for making you confuse here 🙏🏼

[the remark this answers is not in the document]
tednguyendev: Understood! Thanks for the clarification.

Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/116/commits/4c2568f1e4e90ec70f664d467ac38f4f5bc91557) ✋

phisith (approved):

nextjs-13/src/stylesheets/components/_list-newsletter.scss
```diff
@@ -16,11 +16,23 @@
...
 
       color: black;
-
       cursor: pointer;
+      justify-content: space-between;
       background: var(--gray-0);
       border-radius: var(--radius-conditional-2);
       box-shadow: var(--shadow-2);
+
+      .newsletter-name {
```
olivierobert, line 25: There is no need to nest this inside `newsletter-item`. Prefer to keep nesting to a minimum (i.e., you do not need to replicate the HTML structure as it can change):

```css
.list-newsletter {
  .newsletter-wrapper {}
  .newsletter-item {}
  .newsletter-name {}
  ..newsletter-button {}
}

```

> **Note**
> The team prefers using BEM (see https://nimblehq.co/compass/development/code-conventions/css/#class-names). It is not a blocker in this project but this is a heads-up as you will be asked to follow that naming on client project 😉

nextjs-13/src/components/ListNewsletter/index.tsx
```diff
@@ -1,24 +1,27 @@
...
   name: string;
 }
 
 interface Props {
   records: Record[];
+  onAfterCloseCallback: () => undefined;
 }
 
-const ListNewsletter = ({ records }: Props) => {
+const ListNewsletter = ({ records, onAfterCloseCallback }: Props) => {
```
olivierobert, line 13: I understand the need to pass a callback to refresh the list of newsletter (if we were to use Redux or useReducer we would not need to pass down so many props) but this naming is quite confusing as it related to the modal (which is a detail of implementation). So I recommend the following:

```suggestion
const ListNewsletter = ({ records, refreshRecordListCallback }: Props) => {
```

Below in https://github.com/nimblehq/js-framework-benchmark/pull/116/files#diff-223e9b714cf2c9b5f163079f0bd63d3a6eac9c46915fc3afde20a2742897d706R70, the naming makes sense as it related to the modal but not in `ListNewsletter`.

nextjs-13/src/components/ListNewsletter/index.tsx
```diff
@@ -1,24 +1,27 @@
...
-              <li className="newsletter-item" key={item.id}>
-                <span>{item.name}</span>
-              </li>
-            );
-          })}
+          {records.map((item) => (
+            <NewsletterItem
+              key={item.id}
+              item={item}
+              onAfterCloseCallback={onAfterCloseCallback}
```
olivierobert, line 22: Ditto here on the naming:

```suggestion
              onAfterDeleteCallback={refreshRecordListCallback}
```

nextjs-13/src/components/NewsletterItem/index.tsx
```diff
@@ -0,0 +1,61 @@
+import Image from 'next/image';
+import Swal from 'sweetalert2';
```
olivierobert, line 2: As done for Remix (see https://github.com/nimblehq/js-framework-benchmark/blob/develop/remix/app/lib/notification/showNotification.ts), it would be best to abstract the notification implementation (i..e, not having to import `sweetalert2` in more than one place). But if it is too much work we can level it as-is for this project.

olivierobert (changes requested):

olivierobert (approved):

nextjs-13/src/stylesheets/components/_list-newsletter.scss
tednguyendev: 👀 Ohh #til. 

Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/116/commits/92acca69f04d80cc297861f8b3bb60060554ca9f)

nextjs-13/src/components/ListNewsletter/index.tsx
tednguyendev: Yes that's my bad 😭 Thanks for reminding me!

Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/116/commits/88c0be267f1152993e680d082a339f3909afd2d2)

nextjs-13/src/components/ListNewsletter/index.tsx
tednguyendev: Same as the above issue, resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/116/commits/88c0be267f1152993e680d082a339f3909afd2d2)

nextjs-13/src/components/NewsletterItem/index.tsx
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/116/commits/75f2f57cc9ffc0d5e0b786fbb3c3f9c1c1db3586)

olivierobert: It is better but as recommended to Beckham, it would be better to use a more generic name e.g., `alert`, `notification`, etc. We don't need to stick to the library name 😅 What do you think?

tednguyendev: Ah sure! It can be abstracted more 👀 

Improved [here](https://github.com/nimblehq/js-framework-benchmark/pull/116/commits/eb5b553f9b6d154c8e85ca86a43ad30860b4972c)

nextjs-13/src/lib/showAlert/showAlert.ts
```diff
@@ -0,0 +1,12 @@
+import Swal, { SweetAlertIcon } from 'sweetalert2';
+import withReactContent from 'sweetalert2-react-content';
+
+const MySwal = withReactContent(Swal);
+
+export default async function makeSweetAlert(
```
olivierobert, line 6: Did you miss changing the method name here or it is on purpose? 🤔 

Also in terms of directory structure, what do you think of:

```
├── lib
│   └── alert.ts
```

It looks repetitive to have `showAlert/showAlert.ts`.

tednguyendev: That's on purpose. Because I'm seeing the utils on `lib` are usually in folders on `lib`. Seems like the `alert.js` can be on `lib` without on another folder! Resolve [here](https://github.com/nimblehq/js-framework-benchmark/pull/116/commits/b8a09fa9e7118d734f0789147a374367b329cce6)!

PR #117, opened 2023-06-05
[#46] [Backend] As a user I can edit a newsletter

### LINKED ISSUE #46: [Backend] As a user I can edit a newsletter
## Why

Users can edit records of newsletters as needed.

## Acceptance Criteria

- Add a new route to edit newsletters.
- Update the controller and service layers to update newsletters.
- Ensure validation rules are implemented.

tednguyendev:
#46 

## What happened 👀
- Add [`updateNewsletter` to `newsletter.repository.ts`](https://github.com/nimblehq/js-framework-benchmark/pull/117/files#diff-7f62d1d6ff486da9e5be846b0de3dc572e91704e2e913d9aa47ab9e57090f37bR12)
- Add [`PUT` route to newsletter route.js](https://github.com/nimblehq/js-framework-benchmark/pull/117/files#diff-e2c82a3d6cd014d8235827328f63df5db8efb48bbd67175aad18498f1950b96dR38)
- Add tests

## Insight 📝
N/A

## Proof Of Work 📹

Before update:

CleanShot 2023-06-05 at 15 29 04@2x

Update:

CleanShot 2023-06-05 at 15 33 13

After Update:

CleanShot 2023-06-05 at 15 29 59@2x

### DIFF
--- a/nextjs-13/src/app/api/v1/newsletter/[id]/route.test.ts
+++ b/nextjs-13/src/app/api/v1/newsletter/[id]/route.test.ts
@@ -55,47 +55,55 @@ describe('DELETE /v1/newsletter/:id', () => {
 });
 
 describe('PUT /v1/newsletter/:id', () => {
-  const user = { ...userFactory, id: '1' };
-  const newsletter = { id: '2' };
-  const data = {
-    name: 'New name',
-    content: 'New content',
-  };
-  const where = {
-    id: newsletter.id,
-    user: user,
-  };
-  const requestBody = { json: () => data };
-  const args = { where, data };
-
-  beforeEach(() => {
-    appHandler.mockImplementation((_, callback) => callback(user, {}));
-    updateNewsletter.mockResolvedValue({ count: 1 });
-  });
-
   describe('newsletter exists', () => {
     it('updates the newsletter', async () => {
-      const response = await PUT(requestBody, {
-        params: { id: newsletter.id },
-      });
+      const user = { ...userFactory, id: '1' };
+      const newsletter = { id: '2' };
+      const data = {
+        name: 'New name',
+        content: 'New content',
+      };
+      const where = {
+        id: newsletter.id,
+        user: user,
+      };
 
-      expect(updateNewsletter).toHaveBeenCalledWith(args);
+      appHandler.mockImplementation((_, callback) => callback(user, {}));
+      updateNewsletter.mockResolvedValue({ count: 1 });
+
+      const response = await PUT(
+        { json: () => data },
+        { params: { id: newsletter.id } }
+      );
+
+      expect(updateNewsletter).toHaveBeenCalledWith({ where, data });
       expect(response.status).toBe(StatusCodes.OK);
     });
   });
 
   describe('newsletter does NOT exist', () => {
-    beforeEach(() => {
-      updateNewsletter.mockResolvedValue({ count: 0 });
-    });
-
     it('returns error', async () => {
-      const response = await PUT(requestBody, {
-        params: { id: newsletter.id },
-      });
+      const user = { ...userFactory, id: '1' };
+      const newsletter = { id: '2' };
+      const data = {
+        name: 'New name',
+        content: 'New content',
+      };
+      const where = {
+        id: newsletter.id,
+        user: user,
+      };
+
+      appHandler.mockImplementation((_, callback) => callback(user, {}));
+      updateNewsletter.mockResolvedValue({ count: 0 });
+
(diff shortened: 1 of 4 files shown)


nextjs-13/src/app/api/v1/newsletter/[id]/route.ts
```diff
@@ -31,3 +34,43 @@ export async function DELETE(
...
+        id: params.id,
+        user: currentUser,
+      };
+
+      const result = await updateNewsletter({
+        where,
+        data,
+      });
+
+      if (result.count === 0) {
```
Comment (olivierobert) [inline nextjs-13/src/app/api/v1/newsletter/[id]/route.ts:58]: This could be simplified to:

```suggestion
      if (!result.count) {
```

nextjs-13/src/app/api/v1/newsletter/[id]/route.test.ts
```diff
@@ -49,3 +53,53 @@ describe('DELETE /v1/newsletter/:id', () => {
...
+    id: newsletter.id,
+    user: user,
+  };
+  const requestBody = { json: () => data };
+  const args = { where, data };
+
+  beforeEach(() => {
+    appHandler.mockImplementation((_, callback) => callback(user, {}));
+    updateNewsletter.mockResolvedValue({ count: 1 });
+  });
```
Comment (olivierobert) [inline nextjs-13/src/app/api/v1/newsletter/[id]/route.test.ts:74]: For the test above in `DELETE`, the usage of `beforeEach` was more minimal, but this implementation for `PU` is abusing too much on shared fixtures which is an anti-pattern: https://github.com/nimblehq/compass/pull/1274/files#diff-bc1c8aea77bf01da18ac0ba6c224d40694f9c88fa9a6f7d2a574567372ae884cR105. Fixtures should not be shared.

It is better to repeat the fixture creation in each test as 1) it makes the test easier to read, 2) it ensures we only create the required fixtures for each test, and 3) debugging is easier.

So the following is expected:

```javascript
describe('newsletter exists', () => {
    it('updates the newsletter', async () => {
      const user = { ...userFactory, id: '1' };
      const newsletter = { id: '2' };
      const data = {
        name: 'New name',
        content: 'New content',
      };
      const where = {
        id: newsletter.id,
        user: user,
      };

     appHandler.mockImplementation((_, callback) => callback(user, {}));
     updateNewsletter.mockResolvedValue({ count: 1 });

      const response = await PUT({ json: () => data }, {
         params: { id: newsletter.id },
      });

      expect(updateNewsletter).toHaveBeenCalledWith({ where, data });
      expect(response.status).toBe(StatusCodes.OK);
    });
});

describe('newsletter does NOT exist', () => {
   it('returns error', async () => {
      const user = { ...userFactory, id: '1' };
      const newsletter = { id: '2' };

      appHandler.mockImplementation((_, callback) => callback(user, {}));
      updateNewsletter.mockResolvedValue({ count: 0 });

      const response = await PUT({ json: () => {}, {
        params: { id: newsletter.id },
      });
      const responseBody = await response.json();

      expect(response.status).toBe(StatusCodes.UNPROCESSABLE_ENTITY);
      expect(responseBody).toMatchObject({
        message: 'Newsletter could not be updated',
      });
   });
});
```

olivierobert (changes requested):

phisith (approved):
Overall LGTM !!

nextjs-13/src/app/api/v1/newsletter/[id]/route.test.ts
Comment (tednguyendev) [reply nextjs-13/src/app/api/v1/newsletter/[id]/route.test.ts:74]: Ohh I see.

So these fixtures are shared here, instead of with Rspec, the default behavior of factories(using like `let(:user) { User.create }`) are not shared. I thought they also don't share here.

Oke let me improve this.

Comment (tednguyendev) [reply nextjs-13/src/app/api/v1/newsletter/[id]/route.test.ts]: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/117/commits/bfb0be0590523740e6bf06ff7594d02cc26624a1#diff-e9a6bb57657e77990e30f03a0bd834fe36456159e16ea91f33c4bc2de833cf31R58-R109)

nextjs-13/src/app/api/v1/newsletter/[id]/route.ts
Comment (tednguyendev) [reply nextjs-13/src/app/api/v1/newsletter/[id]/route.ts]: Improved [here](https://github.com/nimblehq/js-framework-benchmark/pull/117/commits/bfb0be0590523740e6bf06ff7594d02cc26624a1#diff-e2c82a3d6cd014d8235827328f63df5db8efb48bbd67175aad18498f1950b96dR58)

olivierobert (approved):

PR #113, opened 2023-06-01
[#48] [Backend] As a user I can delete a newsletter

tednguyendev:
#48

## What happened 👀
- Add `deleteNewsletter` to Newsletter Repository
- Add new route to delete a newsletter
- Add tests

## Insight 📝
N/A

### Before(Have 1 newsletter)

CleanShot 2023-06-14 at 14 28 01@2x

### Delete
CleanShot 2023-06-14 at 14 28 49@2x

### After(Have 0 newsletter)
CleanShot 2023-06-14 at 14 29 34@2x

PR #118, opened 2023-06-05
[#47] [Frontend] As a user I can edit a newsletter

### LINKED ISSUE #47: [Frontend] As a user I can edit a newsletter
## Why

Users can edit records of newsletters as needed.

## Acceptance Criteria

- Add a button in the newsletter list to access the page to edit a newsletter.
- Show validation errors in the form to edit a newsletter.

tednguyendev:
#47 

## What happened 👀
- Change `CreateNewsletterModal` into `NewsletterModal`
- Add edit button
- Add the update functions into `CreateOrUpdateNewsletterModal`
- Update `queryNewsletterByUserId`make the query result more consistent
- Add tests

## Insight 📝
`N/A`

## Proof Of Work 📹

https://github.com/nimblehq/js-framework-benchmark/assets/76901736/ee386eb6-9430-45b4-ad55-c936f0e58844


nextjs-13/src/stylesheets/components/_list-newsletter.scss
```diff
@@ -29,6 +29,11 @@
         overflow: hidden;
         text-overflow: ellipsis;
       }
+
+      .icons {
```
olivierobert, line 33: Avoid plurals: https://nimblehq.co/compass/development/code-conventions/#convention-15 In addition, here it is more about having a list of actions that a user can perform:

```
.list-action {
```

nextjs-13/src/components/NewsletterItem/index.tsx
```diff
@@ -38,14 +43,24 @@ const NewsletterItem = ({ item, getData }: Props) => {
...
+          data-testid="btn-edit"
+        />
+        <Image
+          src={`/images/icons/trash.svg`}
+          alt={'Delete button'}
+          width={16}
+          height={16}
+          onClick={handleDelete}
+          data-testid="btn-delete"
+        />
```
olivierobert, line 62: These should be buttons, not images, to be semantically correct.

nextjs-13/src/app/page.tsx
```diff
@@ -31,6 +37,23 @@ const Home = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  async function openCreateModal() {
+    setModalType('create');
+    setIsOpen(true);
+  }
+
+  async function openUpdateModal(item) {
```
olivierobert, line 45: Why isn't this defined in the `ListNewsletter` component? It seems out of scope of the root page 🤔

olivierobert (changes requested):

nextjs-13/src/components/CreateOrUpdateNewsletterModal/index.tsx
```diff
@@ -17,28 +18,61 @@ const customStyles = {
...
+      name: string;
+      content: string;
+    }
+  | undefined;
+
+export type Props = {
   modalIsOpen: boolean;
   setIsOpen: Dispatch<SetStateAction<boolean>>;
+  getData: () => void;
+  modalType: ModalType;
```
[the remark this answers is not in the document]
tednguyendev: Yes that would be a better name for it. I will update and notify you!

nextjs-13/src/app/page.tsx
tednguyendev: Hmm I see.

The reason I do that is that, although it's true defining `openUpdateModal` is not the responsibility of the root page, but without defining `openUpdateModal` here, then instead of we only need to pass one prop, which is the `openUpdateModal` to `ListNewsletter`, we would need to instead pass [3 props into the `ListNewsletter`, which is the `setModalType`, `setIsOpen`, and `setCurrentNewsletter`](https://github.com/nimblehq/js-framework-benchmark/pull/118/files#diff-223e9b714cf2c9b5f163079f0bd63d3a6eac9c46915fc3afde20a2742897d706R46-R55) so that the `ListNewsletter` can create it's own `openUpdateModal`.

So in short, I did that just to make it looks more concise.

What do you think about this? Is passing 3 props instead of 1 is the better practice we prefer taking? 👀

tednguyendev:
@olivierobert Ah, and the PR after the [#43](https://github.com/nimblehq/js-framework-benchmark/pull/110) that I chose to do is the [#48](https://github.com/nimblehq/js-framework-benchmark/pull/113) 🙈. You can use [this filter](https://github.com/nimblehq/js-framework-benchmark/pulls?q=is%3Aopen+is%3Apr+author%3Atednguyendev+sort%3Acreated-asc) so see the order of which PR you can review next.

nextjs-13/src/components/CreateOrUpdateNewsletterModal/index.test.tsx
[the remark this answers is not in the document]
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/118/commits/8561e2ecb72fae90109e2d195ba462c382c47d0d)

nextjs-13/src/components/CreateOrUpdateNewsletterModal/index.tsx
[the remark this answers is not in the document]
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/118/commits/0ab71544b75746c979612bdbe2bd53604feebae1)

nextjs-13/src/components/CreateOrUpdateNewsletterModal/index.tsx
```diff
@@ -47,15 +81,16 @@ const CreateNewsletterModal = ({ modalIsOpen, setIsOpen }: Props) => {
 
       closeModal();
       setLoading(false);
-      toast('Created newsletter success!', 'success');
+      getData();
```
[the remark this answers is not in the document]
tednguyendev: I see 👀 

Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/118/commits/46700231ffdc65fb59dc84d447393fbad449a1dc)!

nextjs-13/src/stylesheets/components/_list-newsletter.scss
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/118/commits/b030f0908e4ebbf81c87e3638f4e04d9c1117ff9)

nextjs-13/src/components/NewsletterItem/index.tsx
tednguyendev: Resolved [here](https://github.com/nimblehq/js-framework-benchmark/pull/118/commits/ce0ea642aa5f42cfb7218eeb1e920c4d7f0431d3)

nextjs-13/src/app/page.tsx
olivierobert: Understood. In this simple case, having the parent owning the state is consistent with [the official best practice](https://react.dev/learn/sharing-state-between-components#a-single-source-of-truth-for-each-state). In more complex applications or use cases, it would be best to use `Context` and `useReducer` (see https://stackoverflow.com/a/63687665)

olivierobert (approved):

PR #120, opened 2023-06-06
[#64] [Frontend] As a user I can create a newsletter

phisith:
[#64]

## What happened 👀

- Implement UI for creating newsletter form
- Add from validation UI
- Implement toast

## Insight 📝

To implement UI and validation there aren't any issues.

On the other side, when Implementing the toast to show the successfully create newsletter. There is a big issue with that. As we use `action` to submit the data to the database, because that `action` is a server function, so we can't call `toast` during the `hydration`. 

To solve this, I pass the toast `message` and `mode` on the `localStorage`, then use `useEffect` to check whether there is any `toast message` or not when the page is mounted, if there is `useEffect` will call `toast`. [Reference](https://www.jacobparis.com/content/remix-form-toast) (* In the reference they are using `Cookie` instead of `localStorage`)

This also allows to call the toast when redirect to another page too. Overall, this requires more setup than normal `React` & `NextJs`, but I'm very happy with the end result.

## Proof Of Work 📹
https://github.com/nimblehq/js-framework-benchmark/assets/59072804/fc35f990-4a7b-4b31-bf68-3ca7023522b2


remix/app/routes/_app.newsletter.create.test.tsx
```diff
@@ -50,14 +48,7 @@ describe('POST /newsletter/create', () => {
...
 
-      expect(await result.json()).toMatchObject<Newsletter>({
-        id: newsletter.id,
-        name: expect.any(String),
-        content: expect.any(String),
-        userId: newsletter.userId,
-        createAt: expect.any(String),
-        updateAt: expect.any(String),
-      });
+      expect(await result.status).toBe(302);
```
tednguyendev, line 51: Friendly reminder: We can use `http-status-codes` here 🙈

remix/app/routes/_app._index.tsx
```diff
@@ -1,12 +1,19 @@
 import { User } from '@prisma/client';
 import { useOutletContext } from '@remix-run/react';
 
+import AppLink from '../components/AppLink';
+
 export default function Index() {
   const [user] = useOutletContext() as [User];
 
   return (
     <>
```
tednguyendev, line 10: I'm just curious but do we need the fragment here? Cause I thought fragments are used to just group childs that not have a parent, but we have a `section` parent here 🤔

remix/app/lib/handler/app.handler.ts
```diff
@@ -1,22 +1,24 @@
...
 export default async function appHandler(
   request: Request,
-  callback: (user: User) => Promise<unknown>
+  callback: (user: User, session: Session) => Promise<unknown>
 ) {
   let userInfo: User;
 
-  const session = (await authenticator.isAuthenticated(request)) as User;
+  const userSession = (await authenticator.isAuthenticated(request)) as User;
+  const cookie = await getSession(request.headers.get('Cookie'));
```
tednguyendev, line 16: Super minor: I think we can move this `cookie` line to after the line:

```typescript
userInfo = await UserRepository.findBy({ email: userSession.email });
```

As we only need to use the `cookie` to pass into the `callback` funfction.

tednguyendev (approved):
LGTM! I like the way you handle `toast` 😄

remix/app/routes/_app._index.tsx
tednguyendev, line 10: I'm just curious but do we need the fragment here? Cause I thought fragments are used to just group childs that not have a parent, but we have a `section` parent here 🤔

tednguyendev: I means the

```typescript
    <>
      <section>
      </section>
    </>
```

remix/app/routes/_app.newsletter.create.test.tsx
phisith: Ohh, I actually have library on the project, but I just follow it 🙊.

I will add it, Thanks for remind me that!

remix/app/routes/_app._index.tsx
phisith: > LGTM! I like the way you handle `toast` 😄

Thanks for that, you can implement the way I do it here as `Remix` and `NextJs` are both `React` base  framework!

remix/app/lib/handler/app.handler.ts
phisith: Yep, That's sound better.

remix/app/routes/_app.newsletter.create.test.tsx
tednguyendev: I know that feeling 😭

remix/app/routes/_app._index.tsx
phisith: For the fragment there, actually this come from my own behavior, as the old version React is define the file as `.js`, so I just find the way to tell which `.js` file is `React` component, by using the `React Fragment` as the mark for me.

Nowadays, we don't actually need that as we define file as `.jsx` or `tsx`

I will remove it, thanks for asking me that, it help me to improve too 😅

remix/app/routes/_app.newsletter.create.test.tsx
phisith: Improved here: [07415ce](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/07415cea75c043a849c912f59b3c4428548ba754)

remix/app/lib/handler/app.handler.ts
phisith: Improved here: https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/07415cea75c043a849c912f59b3c4428548ba754#diff-036147b6dd83af754aaf50a68df29df4853cc9b7360cb958764263be355efd6a

remix/app/routes/_app._index.tsx
phisith: Improved here: [07415ce](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/07415cea75c043a849c912f59b3c4428548ba754)

tednguyendev: Ohh #til 👀

remix/app/components/FormInput/error.tsx
```diff
@@ -0,0 +1,14 @@
+interface ErrorProps {
+  error?: string;
+}
+
+export default function Error({ error }: ErrorProps) {
+  return (
+    <p
```
olivierobert, line 7: Semantically-wise, it might be better to use `div` or `small` tag as `p` tags requires lots of style overrides such as margins

remix/app/components/FormInput/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+jest.mock('remix-validated-form', () => ({
+  useField: jest.fn().mockReturnValue({
+    error: 'Test error',
+    getInputProps: jest.fn().mockReturnValue({}),
+  }),
+}));
+
+describe('FormInput', () => {
+  describe('givens FormInput attribute', () => {
```
olivierobert, line 13: You keep making this typo and here it might be about assigning valid props:

```suggestion
  describe('given valid props', () => {
```

remix/app/components/FormInput/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+      const inputElement = screen.getByPlaceholderText(placeholder);
+      const errorElement = screen.getByText(error);
+
+      expect(labelElement).toBeInTheDocument();
+      expect(inputElement).toBeInTheDocument();
+      expect(errorElement).toBeInTheDocument();
+    });
+  });
+
+  it('enter value', () => {
```
olivierobert, line 32: How about:

```suggestion
  it('sets the text input value', () => {
```

remix/app/components/FormTextarea/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  error?: string;
+}
+
+export default function FormTextarea({ label, name, placeholder }: InputProps) {
+  const { error, getInputProps } = useField(name);
+
+  return (
+    <div className="flex flex-col gap-2 px-0.5">
+      <Label htmlFor={name} label={label} />
+      <textarea
```
olivierobert, line 19: The code for the components of `FormInout` and `FormTextarea` are redundant (and duplicated). How about abstracting this via a prop:

```
<FormTextInput format="short" ... /> -> Renders a text input
<FormTextInput format="long" ... /> -> Renders a textarea
```

remix/app/components/FormInput/error.tsx
olivierobert, line 1: For all components the directory structure must be:

```
ComponentName
│   └── index.tsx
```

So here the directory structure should be:

```
FormInput
├── Error
│   └──  index.tsx
├── Label
│   └──  index.tsx
│   └── index.tsx
```

Refer to https://nimblehq.co/compass/development/code-conventions/javascript/react/#components--screens

remix/app/lib/toast/makeToast.ts
```diff
@@ -0,0 +1,11 @@
+import toast from 'react-hot-toast';
+
+export default function makeToast(toastMode: string, toastMessage: string) {
+  if (toastMode === 'success') {
```
olivierobert, line 4: How about using a switch statement here instead 💡

remix/app/root.tsx
```diff
@@ -8,17 +11,45 @@ import {
...
 export default function App() {
+  const { toastMessage, toastMode } = useLoaderData<typeof loader>();
+
+  useEffect(() => {
+    if (!toastMessage) {
+      return;
+    }
+
+    makeToast(toastMode, toastMessage);
+  }, [toastMessage]);
```
olivierobert, line 51: It might be worth adding [this article as reference](https://www.jacobparis.com/content/remix-form-toast) in the insights section of the pull request as using cookie storage for such use case is unexpected 😅 I am wondering if using localStorage might be better than cookie storage 💡

remix/app/routes/_app.newsletter.create.tsx
```diff
@@ -26,31 +29,43 @@ export async function action({ request }: ActionArgs) {
...
-        <div>
-          <label htmlFor="content">Content</label>
-          <textarea name="content" placeholder="content" />
+    <section className="flex flex-col gap-4">
+      <header className="text-xl font-bold">
+        <h1>Create your Newsletter </h1>
+      </header>
+      <ValidatedForm
+        validator={validator}
+        method="POST"
```
olivierobert, line 56: Does this need to be capitalized? It is not in the `AuthSignInPage` route 🤔 So it looks inconsistent

olivierobert (changes requested):

remix/app/routes/_app.newsletter.create.tsx
phisith: Oh, that is one of my mistakes too, As a newer version of HTML we should use `lowercase` instead of `uppercase`

I will fix it.

 [Reference](https://stackoverflow.com/questions/7790576/what-are-http-method-capitalization-best-practices).

remix/app/root.tsx
phisith: Yep, for that I just forgot to do added the reference that 😅, using `localStorage` might be a better choice here, as we can keep the cookie more secure. Thank you for mentioning that 🙏🏼.

I will improve.

remix/app/lib/toast/makeToast.ts
phisith: Yes, `switch statement` would be better than `if/else` here, but that's I thought my use-case will be only `success`, `error`, and `not provide toastMode`.

But if look at this from the scaling perspective `switch statement` will be better, as we don't use what's gonna come in the future.

I will Improve it.

remix/app/components/FormTextarea/index.tsx
phisith: Yes, that would be much better.

When I'm doing this I also know that those components are duplicated from each other, I am afraid that combining `<input>` and `<textarea>` isn't allowed.

So as suggested, I will improve it.

remix/app/components/FormInput/index.test.tsx
phisith: Yes, that's better

remix/app/components/FormInput/index.test.tsx
phisith: 😅 yes, I will be more careful about it.

remix/app/components/FormInput/error.tsx
phisith: Yep will prove that.

phisith: Improve here: [0af22fb](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/0af22fb487b5618b545e2edf90f9b9e0fe94da67)

remix/app/components/FormInput/index.test.tsx
phisith: Improved here: [0af22fb](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/0af22fb487b5618b545e2edf90f9b9e0fe94da67)

remix/app/components/FormInput/index.test.tsx
phisith: Improved here: [0af22fb](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/0af22fb487b5618b545e2edf90f9b9e0fe94da67)

remix/app/components/FormTextarea/index.tsx
phisith: Improved here: [0af22fb](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/0af22fb487b5618b545e2edf90f9b9e0fe94da67)

remix/app/routes/_app.newsletter.create.tsx
phisith: Fixed here: [08ab23a](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/08ab23adcb5bda8ceb549edbbc9a3671d2288375)

remix/app/lib/toast/makeToast.ts
phisith: Improved here: [7ba7a3b](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/7ba7a3b2c928919163e49e2c82434c454d3bbe95)

remix/app/lib/toast/makeToast.ts
```diff
@@ -0,0 +1,15 @@
+import toast from 'react-hot-toast';
+
+export default function makeToast(
+  toastMode: string | null,
+  toastMessage: string | null
```
olivierobert, line 5: Don't you think we could pass an object instead of having two attributes:

```suggestion
export default function showNotification({
  text: string
  type?: string,
}  : Notification)
```

> **Note**
> The attribute text should not be nullable IMO. The interface toastItems (renamed to Notification) could be define here instead.

olivierobert (commented):
Some additional minor improvements

remix/app/lib/handler/app.handler.test.ts
```diff
@@ -17,10 +17,14 @@ jest.mock('../../config/auth.server', () => ({
   },
 }));
 
-const mockRequest = {};
-const mockCallback = jest.fn();
-
 describe('appHandler', () => {
+  const mockRequest: any = {
+    headers: {
+      get: jest.fn(() => 'Cookie Value'),
```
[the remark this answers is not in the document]
phisith: In the last implementation, I changed the toast to use `localStorage` instead of the `Cookie`, and the user is still on the `Cookie`.

For that one, I forgot the remove it.

I will remove it.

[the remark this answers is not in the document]
phisith: Removed here: [f078d2a](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/f078d2aa6e5b90e3f455452ffa7e67ef3fd64881)

remix/app/helpers/localStorage.helper.ts
```diff
@@ -0,0 +1,16 @@
+export function setToastItems(toastMode: string, toastMessage: string) {
```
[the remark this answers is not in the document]
phisith: Improved here: [f078d2a](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/f078d2aa6e5b90e3f455452ffa7e67ef3fd64881)

remix/app/helpers/localStorage.helper.ts
```diff
@@ -0,0 +1,16 @@
...
+  localStorage.setItem('toastMode', toastMode);
+  localStorage.setItem('toastMessage', toastMessage);
+}
+
+export function getToastItems() {
```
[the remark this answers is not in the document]
phisith: Improved here: [f078d2a](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/f078d2aa6e5b90e3f455452ffa7e67ef3fd64881)

remix/app/lib/toast/makeToast.ts
phisith: Yep, Passing an object would be better than two attributes. Thank you for mentioning that 😅 🙏🏼 

Improved: [f078d2a](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/f078d2aa6e5b90e3f455452ffa7e67ef3fd64881)

remix/app/helpers/localStorage.helper.ts
```diff
@@ -0,0 +1,18 @@
+import { Notification } from '../types';
+
+export function addNotification({ text, type }: Notification) {
+  localStorage.setItem('notificationText', text);
+  localStorage.setItem('notificationType', type as string);
```
olivierobert, line 5: Based on the comment to refactor using an object (https://github.com/nimblehq/js-framework-benchmark/pull/120#discussion_r1226039076), I expected you to revisit the LocalStorage implementation, too 😅 

```suggestion
  localStorage.setItem('notification', { text, type });
```

remix/app/helpers/localStorage.helper.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export function getNotificationList() {
```
olivierobert, line 8: Unless we decide to store more than one object of notifications (so an array of notificaitions), we could rename it to:

```suggestion
export function getNotification() {
```

remix/app/helpers/localStorage.helper.ts
```diff
@@ -0,0 +1,18 @@
...
+  const text = localStorage.getItem('notificationText');
+  const type = localStorage.getItem('notificationType');
+
+  return { text, type };
+}
+
+export function removeNotificationList() {
```
olivierobert, line 15: Ditto here:

```suggestion
export function removeNotification() {
```

olivierobert (changes requested):
It is good to be fast in integrating code review comments, so you need to take more time to think through the changes 🙈 One change might require further changes that you should think about. These should be the final changes 🙏

remix/app/helpers/localStorage.helper.ts
phisith: Improved here: [5b3237a](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/5b3237af21e9837710fef39f565836071a15e44b)

remix/app/helpers/localStorage.helper.ts
phisith: Improved here: [5b3237a](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/5b3237af21e9837710fef39f565836071a15e44b)

remix/app/helpers/localStorage.helper.ts
phisith: For this one, I know that storing the object on the `localStorage` use `JSON.stringify()`, and `JSON parse` for retrieving the data. That's why I think storing `key`+`value` one by one might be simpler.

But in terms of structuring the `localStorage` by wrapping the object into one will be a better approach.

Improved here: [5b3237a](https://github.com/nimblehq/js-framework-benchmark/pull/120/commits/5b3237af21e9837710fef39f565836071a15e44b)

phisith:
> It is good to be fast in integrating code review comments, so you need to take more time to think through the changes 🙈 One change might require further changes that you should think about. These should be the final changes 🙏

Yep, so next time I will slow things down a bit and think about the change in detail 🙏🏼.

For my personal, sometimes I also don't know which way will work best for us, I think this will take me some time to learn things from the higher level dev than me 😅.

olivierobert (approved):

PR #126, opened 2023-06-08
[#65] [Backend] As a user I can edit a newsletter 

phisith:
[#65]

## What happened 👀

- Create an `update` method on `newsletter.repository` to edit the newsletter.
- Create a new route for updating the newsletter.
- Implement update rule and form validation on update newsletter `action`.
- Add testing.

## Insight 📝

- N/A

#### • before the update
Screenshot 2566-06-09 at 16 33 32

#### • update (no styling)
Screenshot 2566-06-09 at 16 34 04

#### • after the update
Screenshot 2566-06-09 at 16 34 29

### DIFF
--- a/remix/app/repositories/newsletter.server.test.ts
+++ b/remix/app/repositories/newsletter.server.test.ts
@@ -26,4 +26,31 @@ describe('Newsletter Repository', () => {
       );
     });
   });
+
+  describe('findOne', () => {
+    it('returns a newsletter', async () => {
+      const newsletter = { ...newsletterFactory };
+
+      prismaMock.newsletter.findUnique.mockResolvedValue(newsletter);
+
+      await expect(
+        NewsletterRepository.findOne({ where: { id: newsletter.id } })
+      ).resolves.toEqual(newsletter);
+    });
+  });
+
+  describe('update', () => {
+    it('updates a newsletter', async () => {
+      const newsletter = { ...newsletterFactory };
+
+      prismaMock.newsletter.update.mockResolvedValue(newsletter);
+
+      await expect(
+        NewsletterRepository.update({
+          where: { id: newsletter.id },
+          data: newsletter,
+        })
+      ).resolves.toEqual(newsletter);
+    });
+  });
 });
--- a/remix/app/repositories/newsletter.server.test.ts
+++ b/remix/app/repositories/newsletter.server.test.ts
@@ -43,14 +43,14 @@ describe('Newsletter Repository', () => {
     it('updates a newsletter', async () => {
       const newsletter = { ...newsletterFactory };
 
-      prismaMock.newsletter.update.mockResolvedValue(newsletter);
+      prismaMock.newsletter.updateMany.mockResolvedValue({ count: 1 });
 
       await expect(
         NewsletterRepository.update({
           where: { id: newsletter.id },
           data: newsletter,
         })
-      ).resolves.toEqual(newsletter);
+      ).resolves.toEqual({ count: 1 });
     });
   });
 });
--- a/remix/app/routes/_app.newsletter.update.$id.test.tsx
+++ b/remix/app/routes/_app.newsletter.update.$id.test.tsx
@@ -4,7 +4,7 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { StatusCodes } from 'http-status-codes';
+import { Newsletter } from '@prisma/client';
 
 import { action } from './_app.newsletter.update.$id';
 import { authenticator } from '../config/auth.server';
@@ -25,7 +25,7 @@ describe('POST /newsletter/update/:id', () => {
   });
 
   describe('givens valid newsletter params with a creator user', () => {
-    it('redirect to main page, when a newsletter is successfully updated', async () => {
+    it('update a newsletter', async () => {
       const userAttributes = { id: '1' };
       const newsletterAttributes = { userId: '1' };
       const user = { ...userFactory, ...userAttributes };
@@ -43,7 +43,7 @@ describe('POST /newsletter/update/:id', () => {
 
       const request = makeRequest({
         url: `/newsletter/update/${newsletter.id}`,
-        method: 'POST',
+        method: 'post',
         body,
       });
 
@@ -53,8 +53,14 @@ describe('POST /newsletter/update/:id', () => {
         context: {},
       });
 
-      expect(await result.status).toBe(StatusCodes.MOVED_TEMPORARILY);
-      expect(await result.headers.get('Location')).toBe('/');
+      expect(await result.json()).toMatchObject<Newsletter>({
+        id: newsletter.id,
+        name: expect.any(String),
+        content: expect.any(String),
+        userId: newsletter.userId,
+        createAt: expect.any(String),
+        updateAt: expect.any(String),
+      });
     });
   });
 
@@ -73,7 +79,7 @@ describe('POST /newsletter/update/:id', () => {
 
       const request = makeRequest({
         url: `/newsletter/update/${newsletter.id}`,
-        method: 'POST',
+        method: 'post',
         body,
       });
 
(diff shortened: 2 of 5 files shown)


remix/app/routes/_app.newsletter.update.$id.tsx
```diff
@@ -0,0 +1,88 @@
...
+export async function action({ request, params }: ActionArgs) {
+  return appHandler(request, async (user) => {
+    const result = await validator.validate(await request.formData());
+
+    if (result.error) {
+      return validationError(result.error);
+    }
+
+    const newsletterExists = await NewsletterRepository.findOne({
+      where: { id: params.id },
```
olivierobert, line 29: The query should also inlcude the `user.id` for simplicity. If none is found it should redirect the user back to the 404 page. There is no need to show an authorization error (https://github.com/nimblehq/js-framework-benchmark/pull/126/files#diff-d71d009bbf395afa6601247a25a1001d022a062369df91b53b77d2b6713f086dR40) as the use case is simple.

remix/app/routes/_app.newsletter.update.$id.tsx
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    if (!newsletterExists) {
+      return validationError({
```
olivierobert, line 33: We should instead return a 404 page: https://remix.run/docs/en/main/guides/not-found#how-to-send-a-404. Landing on a page with an error message is not great 😅

remix/app/routes/_app.newsletter.update.$id.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+}));
+
+jest.mock('../lib/handler/app.handler');
+
+describe('PUT /newsletter/update/:id', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('givens valid newsletter params with a creator user', () => {
```
olivierobert, line 29: You still make the same typo 🙈 

```suggestion
  describe('given valid newsletter params', () => {
```

We should also wrap the block into a parent block for authenticated users since `callback(user)` is implemented for the  `appHandler`:

```javascript
describe('given the user is authnticated () => {
  describe('given valid newsletter params () => {
  })
})

describe('given the user is NOT authnticated () => {

})
```

remix/app/routes/_app.newsletter.update.$id.tsx
```diff
@@ -0,0 +1,88 @@
...
+        fieldErrors: {
+          Error: 'Newsletter not found',
+        },
+      });
+    }
+
+    if (newsletterExists.userId !== user.id) {
```
tednguyendev, line 40: Or you can use use [Prisma's findMany](https://www.prisma.io/docs/concepts/components/prisma-client/crud#get-all-records), which will also allow you to pass the `user.id` to the query and we will be able to check both things(the newsletter id exists, and the user of the newsletter is the current user) 👀

olivierobert (changes requested):

remix/app/routes/_app.newsletter.update.$id.tsx
```diff
@@ -0,0 +1,88 @@
...
+      return validationError({
+        fieldErrors: {
+          Error: 'You are not allowed to update this newsletter',
+        },
+      });
+    }
+
+    const newsletterUpdated = await NewsletterRepository.update({
```
tednguyendev, line 48: Or you can use [Prisma's updateMany](https://www.prisma.io/docs/concepts/components/prisma-client/crud#update-multiple-records), which will allow us to do all 3 things(check newsletter id exists, check if the user of the newsletter is the current user, and update the newsletter) at once with only 1 query 🤣 In case you want to give it a try, here is how [I used it](https://github.com/nimblehq/js-framework-benchmark/pull/117/files#diff-e2c82a3d6cd014d8235827328f63df5db8efb48bbd67175aad18498f1950b96dR53-R63).

phisith: Yep, that's a good option I migrated to that 😁, Thanks for the suggestion.

Improved here: [d426473](https://github.com/nimblehq/js-framework-benchmark/pull/126/commits/d426473d63ea5e96d4c67a790b23169b39f78847)

remix/app/routes/_app.newsletter.update.$id.tsx
phisith: Well noted.

Improved here: [d426473](https://github.com/nimblehq/js-framework-benchmark/pull/126/commits/d426473d63ea5e96d4c67a790b23169b39f78847)

remix/app/routes/_app.newsletter.update.$id.tsx
phisith: Yep! I have improved here: [d426473](https://github.com/nimblehq/js-framework-benchmark/pull/126/commits/d426473d63ea5e96d4c67a790b23169b39f78847)

remix/app/routes/_app.newsletter.update.$id.test.tsx
phisith: Umm, Sorry if I keep making the typo on testing, I know that we should write one test to expect one behavior only. In this case, I just want to provide more context on the description. 

But I know my test description still needs to improve, do you have any good source that shows how to write the test description?

Improved here: [d426473](https://github.com/nimblehq/js-framework-benchmark/pull/126/commits/d426473d63ea5e96d4c67a790b23169b39f78847)

olivierobert: We don't have specific content on Compass for it yet (we should). So the cloest I can recommend is https://www.betterspecs.org/

remix/app/repositories/newsletter.server.test.ts
```diff
@@ -26,4 +26,32 @@ describe('Newsletter Repository', () => {
...
+
+      await expect(
+        NewsletterRepository.findOne({ where: { id: newsletter.id } })
+      ).resolves.toEqual(newsletter);
+    });
+  });
+
+  describe('update', () => {
+    it('updates a newsletter', async () => {
+      const result = { count: 1 };
```
olivierobert, line 44: That variable is not really needed. Tests follow the AAA pattern (Arrange - Act - Assert) so since it will be used for assertion it should not be in the first block `Arrange`. 

Instead do the following as it is also more readable (no need to scroll back up 🙈 ):

```javascript
     await expect(
        NewsletterRepository.update({
          where: { id: newsletter.id },
          data: newsletter,
        })
      ).resolves.toEqual({ count: 1 });
```

olivierobert (changes requested):
Just one minor fix to apply

remix/app/routes/_app.newsletter.update.$id.test.tsx
phisith: Thanks for the recommendation @olivierobert, will try to follow the practices from that website first.

remix/app/repositories/newsletter.server.test.ts
phisith: Well noted, Improved here: [7ad4f35](https://github.com/nimblehq/js-framework-benchmark/pull/126/commits/7ad4f356021b07ac686d227f05e4dc7e12b20884)

olivierobert, line 44: You forgot to remove it following the last update 🙃

phisith: Oh, for that one I keep it for the mock update on line 47 😬 or I should write `{count: 1}` on the mock too.

olivierobert: Yes. It is okay to repeat code in tests. It is part of the best practice not to not over-engineering tests (see https://github.com/nimblehq/compass/pull/1274/files#diff-bc1c8aea77bf01da18ac0ba6c224d40694f9c88fa9a6f7d2a574567372ae884cR101)

phisith: Yep, Updated here: [44c2663](https://github.com/nimblehq/js-framework-benchmark/pull/126/commits/44c2663cca727141d796f419d81b3b53cc5e7786)

PR #129, opened 2023-06-09
Refactor API endpoints into Server Actions

tednguyendev:
https://github.com/nimblehq/js-framework-benchmark/pull/106#discussion_r1220630242

## What happened 👀
- Add [the new actions.tsx file](https://github.com/nimblehq/js-framework-benchmark/pull/129/files#diff-ea081786071108e74c9128ec18214e4b25464c487e41f4f3cf2961361d3a9b8a) to contains all actions for 4 Server Actions(Create, Update, Delete, Send)
- Modify client code to use Server Actions instead of calling APIs
- Add and modify tests
- Remove all the API codes and tests of 4 actions

### Purpose of PR
As discussed [here](https://github.com/nimblehq/js-framework-benchmark/pull/106#discussion_r1220630242), this is the PR to refactor all the API code that can be replaced by [Next.js Server Actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions).

### API to Server Actions
It's surprising for me to see that it's very easy to refactor a flow that's using API to using Server Action in Next.js v13.

For example, for this sending newsletter feature, to use Server Action, based [on the refactored code here](https://github.com/nimblehq/js-framework-benchmark/pull/129/files#diff-1410c2c61c8257ce65caa55dcf666a7b6315edc7ddd9c166d4b608a8fe552feb), we just need to:
- Replace the `axios` call without our server action(the function `sendNewsletter` in this example)
- Wrap the whole calling server action thing in [`startTransition`](https://github.com/nimblehq/js-framework-benchmark/pull/129/files#diff-1410c2c61c8257ce65caa55dcf666a7b6315edc7ddd9c166d4b608a8fe552febR50)(`startTransition` is 1 of the 3 ways to use Server Action in Nextjs, along with [`action`](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions#action) and [`formAction`](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions#formaction))
- Everything else stays the same

### Cons
It is easy to use this new Server Action feature, but since this is still a new feature, there are some difficulties I faced while trying it, such as:
- [Not very compatible with `NextAuth`](https://github.com/nextauthjs/next-auth/issues/7486)(when try to get user session on server side, NextAuth raise errors), but luckily, there is [this work around](https://github.com/nextauthjs/next-auth/issues/7486#issuecomment-1543747325)
- When we run tests, it raises some [import errors](https://github.com/vercel/next.js/issues/47448#issuecomment-1483028284), resulting in having to do a lot of import mock like [this](https://github.com/nimblehq/js-framework-benchmark/pull/129/files#diff-4b9ae30a4e34a98e0e0d12af3e4872e7ac9865c68f016c1e6667d4336ce08ce1R16-R21) to make it works.
- It also raises errors when running tests if we import server actions from a file that has "use server" to test files. So instead of only having 2 file, `actions/newsletter.tsx`, I have to create a new [`actions/actions.tsx`](https://github.com/nimblehq/js-framework-benchmark/pull/129/files#diff-ea081786071108e74c9128ec18214e4b25464c487e41f4f3cf2961361d3a9b8a) file that contains the server actions but not the "use server", and when I need to use the server actions on client components, I use it through this [`actions/newsletter.tsx`](https://github.com/nimblehq/js-framework-benchmark/pull/129/files#diff-c4535b6fc32b017b9c90ca00f3e5baaa01f19f24913e5725a69319a7376c512a)(which only to import actions from `actions.tsx` AND ADD the "use server").

### Refs
- This is a very easy to understand introduction about Server Actions: [Next.js Server Actions... 5 awesome things you can do](https://www.youtube.com/watch?v=O94ESaJtHtM&ab_channel=BeyondFireship&themeRefresh=1)

## Proof Of Work 📹
`N/A`


nextjs-13/src/lib/request/getServerSession.ts
```diff
@@ -0,0 +1,19 @@
+import { cookies, headers } from 'next/headers';
+import { getServerSession as originalGetServerSession } from 'next-auth';
+
+import { authOptions } from 'config/auth';
+
```
olivierobert, line 5: Let's add a comment mentioning that it is a workaround and add a link to the source (https://github.com/nextauthjs/next-auth/issues/7486#issuecomment-1543747325)

olivierobert (approved):
Just one minor comment but this looks good. 

I am unsure about grouping all actions into the same file (`actions.tsx`) if the application was larger  but the decoupling into a separate directory is a great idea. It looks similar to Redux.

nextjs-13/src/lib/request/getServerSession.ts
tednguyendev: Added [here](https://github.com/nimblehq/js-framework-benchmark/pull/129/commits/64559b097e6d1a108ad3691bd16bb66eeb47a5ca)! 🙈

PR #130, opened 2023-06-09
[#66] [Frontend] As a user I can edit a newsletter

phisith:
[#66]

## What happened 👀

- Implement `UI` for editing the newsletter page
- Add an edit button on the `newsletterList`
- Implement changing button color function, **using in `AppLink` right now**
- Add testing

## Insight 📝

- N/A

## Proof Of Work 📹
https://github.com/nimblehq/js-framework-benchmark/assets/59072804/587b4d19-fb75-474b-a134-f143e4fbff4a


remix/app/routes/_app.newsletter.update.$id.tsx
```diff
@@ -58,19 +78,31 @@ export default function Index() {
...
+        <h1>Update your Newsletter </h1>
+      </header>
+      <ValidatedForm
+        validator={validator}
+        defaultValues={newsletter}
+        method="put"
+      >
+        <div className="flex flex-col gap-6">
+          <FormInput
+            label={'Name'}
```
tednguyendev, line 92: Can we just remove all brackets like this 👀 :

```typescript
  <FormInput
    label="Content"
    name="content"
    placeholder="content"
    format="long"
  />
```

remix/app/routes/_app.newsletter.update.$id.tsx
```diff
@@ -58,19 +78,31 @@ export default function Index() {
...
+            name={'name'}
+            placeholder={'name'}
```
tednguyendev, line 94: I think it would be a bit better if the `placeholder` can be capitalized like "Name" or "Please input the newsletter name here" :D

remix/app/routes/_app.newsletter.update.$id.tsx
```diff
@@ -42,11 +44,29 @@ export async function action({ request, params }: ActionArgs) {
   });
 }
 
+export const loader = async ({ params }: ActionArgs) => {
+  const newsletter = await NewsletterRepository.findOne({
+    where: { id: params.id },
```
tednguyendev, line 49: I think we should also pass the `userId` into this query, right? 🤔

remix/app/helpers/button.helper.ts
```diff
@@ -0,0 +1,33 @@
+export function setButtonColor(color?: string) {
```
tednguyendev, line 1: I think this is more like a method to get something(a string that contains tailwind classes), more than a method to set something 😸 

Maybe `getButtonClassNames`? 🙈

phisith: Yep, but I think `getButtonClassNames` is too general for me like I just want to get the color from the method, should we name it `getButtonColorClassNames` or `getButtonStyle`. 

What do you think is the best here, I'm Ok if you go with `getButtonClassNames` I can change it to. 😂

remix/app/routes/_app.newsletter.update.$id.tsx
phisith: Umm, that's my bad I missed there 🙏🏼

remix/app/helpers/button.helper.ts
tednguyendev: I see. The `getButtonColorClassNames` is a bit long but it sounds best to me 😄

remix/app/routes/_app.newsletter.update.$id.tsx
phisith: Yep, I just realize that for the `normal string` props that are not a variable, it is better to use it without `{}` which helps use to virtualize which one is just a `string` or `variable` that passes to the component.

Good suggestion @tednguyendev  💪🏽

remix/app/routes/_app.newsletter.update.$id.tsx
phisith: Improved here: [b4eddfe](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/b4eddfee4c7915ecdf0fca12edc7b25d45db97d8)

remix/app/routes/_app.newsletter.update.$id.tsx
phisith: Added here: [b4eddfe](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/b4eddfee4c7915ecdf0fca12edc7b25d45db97d8)

remix/app/helpers/button.helper.ts
phisith: ya I think so, how about `getButtonStyle` kinda like ok right? properly not the good

tednguyendev: It also sounds ok! As long as it start with `get`, I think its better 😄

olivierobert: 1. The switch is not really on the color but on the type. Yes it will change the color styles but from a developer point of view, it makes more sense to have `setButtonType('primary')`.
2. It might be best to colocate this code in the directory `AppLink` to keep all related code together 💡 I doubt that this code will be used somewhere else than in `AppLink`

remix/app/routes/_app.newsletter.update.$id.tsx
tednguyendev: It's better now! Just curious, why did you also change the method's name to `findFirst`? 👀

remix/app/components/AppLink/index.tsx
```diff
@@ -1,15 +1,20 @@
...
+import { setButtonColor } from '../../helpers/button.helper';
+
 interface ButtonProps {
   name: string;
   href: string;
+  color?: string;
 }
 
-export default function AppLink({ name, href }: ButtonProps) {
+export default function AppLink({ name, href, color }: ButtonProps) {
```
olivierobert, line 11: Shouldn't this component be called `Button` instead of `AppLink`? There are button props and helpers 😅

remix/app/repositories/newsletter.server.test.ts
```diff
@@ -27,14 +27,14 @@ describe('Newsletter Repository', () => {
     });
   });
 
-  describe('findOne', () => {
+  describe('findFirst', () => {
```
olivierobert, line 30: Why the change from `findOne` to `findFirst`?

remix/app/routes/_app.newsletter.update.$id.test.tsx
```diff
@@ -35,65 +36,112 @@ describe('PUT /newsletter/update/:id', () => {
...
+
+          expect(await result.json()).toMatchObject({
+            name: expect.any(String),
+            content: expect.any(String),
           });
-        }
+        });
+      });
+
+      describe('given an invalid newsletter params', () => {
```
olivierobert, line 117: Syntax issue:

```suggestion
      describe('given invalid newsletter params', () => {
```

Since it is plural, we cannot use `an`

remix/app/routes/_app.newsletter.update.$id.test.tsx
```diff
@@ -35,65 +36,112 @@ describe('PUT /newsletter/update/:id', () => {
...
       });
     });
+  });
 
-    describe('given an invalid newsletter params', () => {
-      it('returns error NOT FOUND', async () => {
-        const newsletter = { ...newsletterFactory };
+  describe('action', () => {
+    beforeEach(() => {
+      jest.clearAllMocks();
```
olivierobert, line 70: Shouldn't this moved to `afterEach` (like a teardown) in the first `describe` block?

```
describe('Newsletter update', () => {
   afterEach(() => {
      jest.clearAllMocks();
  });

  describe('loader', () => {
}
```

remix/app/routes/_app.newsletter.update.$id.test.tsx
```diff
@@ -21,12 +22,12 @@ jest.mock('../config/auth.server', () => ({
 
 jest.mock('../lib/handler/app.handler');
 
-describe('PUT /newsletter/update/:id', () => {
-  beforeEach(() => {
-    jest.clearAllMocks();
-  });
+describe('Newsletter update', () => {
+  describe('loader', () => {
```
olivierobert, line 26: The previous `describe` block was better. I recommend the following structure:

```javascript
describe('given the user is authenticated', () => {
  describe('given valid newsletter params', () => {
    it('loads the newsletter details', async () => {
      //...
    });
  });

  describe('given valid newsletter updated params', () => {
    it('updates the newsletter', async () => {
      //...
    });
  });

   // nice to have test but not compulsory IMO
   describe('given INVALID newsletter updated params', () => {
    it('returns an error', async () => {
      //...
    });
  });
});

describe('given the user is NOT authenticated', () => {
  it('returns an error', async () => {

  });
})
```

remix/app/routes/_app.newsletter.update.$id.tsx
phisith: Since I already have `findMany`, I think is better to name it `findFirst` too, I think is better to clarify the behavior of the method from the name too.

remix/app/routes/_app.newsletter.update.$id.tsx
```diff
@@ -42,11 +44,31 @@ export async function action({ request, params }: ActionArgs) {
...
+  });
+};
+
 export default function Index() {
   const result = useActionData();
+  const newsletter = useLoaderData();
   const navigate = useNavigate();
 
   useEffect(() => {
+    if (Object.keys(newsletter).length === 0) {
```
olivierobert, line 63: Why do we need to check on the object keys? Is not the following sufficient?

```javascript
if (!newsletter) {

}
```

olivierobert (changes requested):

remix/app/helpers/button.helper.ts
phisith: Well noted 👍🏼

remix/app/components/AppLink/index.tsx
phisith: Oh, that's my bad 🙊, this is my typo on defining the props name, this should be `AppLinkProps` and this is the Link button element that uses for navigation.

remix/app/helpers/button.helper.ts
phisith: Improved here [b357dd3](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/b357dd39fc2364fe1acf62523fa9f7463e88caed)

remix/app/components/AppLink/index.tsx
phisith: Improved here [b357dd3](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/b357dd39fc2364fe1acf62523fa9f7463e88caed)

remix/app/repositories/newsletter.server.test.ts
phisith: For this, I was thinking to name the same `Prisma` method and try to make it more meaningful, but thinking it again they both return a similar result. So I change it back now.

remix/app/routes/_app.newsletter.update.$id.test.tsx
phisith: Yep Fixed here: [b357dd3](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/b357dd39fc2364fe1acf62523fa9f7463e88caed)

remix/app/routes/_app.newsletter.update.$id.test.tsx
phisith: Fixed here: [b357dd3](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/b357dd39fc2364fe1acf62523fa9f7463e88caed)

remix/app/routes/_app.newsletter.update.$id.test.tsx
phisith: Roger that, first I thought separate `action` and `loader` will be good. At the end it cause some duplicate code there.

Improved here: [b357dd3](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/b357dd39fc2364fe1acf62523fa9f7463e88caed)

remix/app/routes/_app.newsletter.update.$id.tsx
phisith: Umm, this was my bad I was over engineer this.

Fixed here: [b357dd3](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/b357dd39fc2364fe1acf62523fa9f7463e88caed)

remix/app/helpers/button.helper.ts
phisith: is it ok, if I still want to separate it from the `AppLink` directory, like this method might be used for another similar button component in the future?

olivierobert: > this method might be used for another similar button component in the future

Well, it is not now 😅 So it can be moved once that case happens. Prefer not to over-optimize for unknown future needs. The codebase should reflect the current and future confirmed needs.

phisith: Solved here: [67a8330](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/67a83308dcdfa4aa48b422d8e184137f5db93c77)

remix/app/components/AppLink/helper.ts
```diff
@@ -0,0 +1,33 @@
+export function setButtonType(color?: string) {
```
olivierobert, line 1: With this update, shouldn't the param name be changed too?

```suggestion
export function setButtonType(type?: string) {
```

remix/app/components/AppLink/index.tsx
```diff
@@ -1,15 +1,20 @@
...
+import { setButtonType } from './helper';
+
+interface AppLinkProps {
   name: string;
   href: string;
+  color?: string;
 }
 
-export default function AppLink({ name, href }: ButtonProps) {
+export default function AppLink({ name, href, color }: AppLinkProps) {
```
olivierobert, line 11: Ditty here on switching to `type`

olivierobert (commented):
Just some minor fixes and this can be merged

remix/app/components/AppLink/helper.ts
phisith: Updated here: [9b0ebe5](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/9b0ebe5ec2af4123c6639502977bb5a584cbc878)

remix/app/components/AppLink/index.tsx
phisith: Updated here: [9b0ebe5](https://github.com/nimblehq/js-framework-benchmark/pull/130/commits/9b0ebe5ec2af4123c6639502977bb5a584cbc878)

olivierobert (approved):

olivierobert:
@tednguyendev your review is required so that it can be merged

tednguyendev (approved):

tednguyendev:
> @tednguyendev your review is required so that it can be merged

Done!

